PR #1, opened 2025-04-18
docs: Add readthedocs link to pyproject/PyPI

DeflateAwning:
# Motivation

Make the entry on PyPI look better and have all the relevant links. 

I'm excited to try this project out more!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,8 @@ requires-python = ">=3.11"
 version = "0.0.0"
 
 [project.urls]
-repository = "https://github.com/quantco/dataframely"
+Repository = "https://github.com/quantco/dataframely"
+Documentation = "https://dataframely.readthedocs.io/"
 
 [tool.maturin]
 module-name = "dataframely._extre"


pyproject.toml
```diff
@@ -22,7 +22,8 @@ requires-python = ">=3.11"
 version = "0.0.0"
 
 [project.urls]
-repository = "https://github.com/quantco/dataframely"
+Repository = "https://github.com/quantco/dataframely"
+Documentation = "https://dataframely.readthedocs.io/"
```
borchero, line 26: The [documentation](https://packaging.python.org/en/latest/specifications/well-known-project-urls/#well-known-labels) lists the well-known labels as lowercase, I guess this would be more fitting:

```suggestion
repository = "https://github.com/quantco/dataframely"
documentation = "https://dataframely.readthedocs.io/"
```

borchero (commented):
Thanks for improving the documentation! :)

pyproject.toml
DeflateAwning: On pypi, it's shown verbatim as written in that key. Your repo is the only one I've ever seen where those are written as lowercase. Looks out-of-place to me.

Not sure why they're lowercase in the docs there - super weird. Kinda seems like the pyproject docs suggest they're case-insensitive (and thus lowercase is the default), while PyPI (the website) interprets them as case-sensitive.

I stand by keeping them titlecase

borchero (approved):

pyproject.toml
DeflateAwning: Thanks :)

PR #5, opened 2025-04-19
feat: Allow to inline collection member columns for sampling

borchero:
# Motivation

When providing overrides for column values in `Collection.sample`, it would be nice to optionally inline columns of members. What do I mean by that? Imagine that you have two members

```python
class A(dy.Schema):
    x = dy.Int16(primary_key=True)
    y = dy.Int16()

class B(dy.Schema):
    x = dy.Int16(primary_key=True)
    z = dy.Int16()
```

in a collection

```
class C(dy.Collection):
    a: dy.LazyFrame[A]
    b: dy.LazyFrame[B]
```

For sampling you can currently do

```python
C.sample(overrides=[{"a": {"y": 5}}, "b": {"z": 6}}])
```

However, it would be nice to "inline" the column values since it is clear what member `y` and `z` belong to, i.e. do

```python
C.sample(overrides=[{"y": 5, "z": 6}])
```

This significantly reduces the amount of typing required to use overrides.

# Changes

- Introduce a new option `inline_for_sampling: bool` for the `dy.CollectionMember` annotation which can be used to inline the columns of a collection member
- Add checks that inlining a collection member is valid (it does not work if either (1) the member is a 1:N relation, i.e. does not share the common primary key, or (2) it isn't clear which member an inlined column value would refer to)

### DIFF
--- a/tests/collection/test_sample.py
+++ b/tests/collection/test_sample.py
@@ -6,7 +6,10 @@ from typing import Annotated, Any
 import pytest
 
 import dataframely as dy
+from dataframely.exc import ImplementationError
 from dataframely.random import Generator
+from dataframely.testing import create_collection
+from dataframely.testing.factory import create_collection_raw
 
 
 class MyFirstSchema(dy.Schema):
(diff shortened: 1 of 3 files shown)


tests/collection/test_sample.py
```diff
@@ -100,6 +117,21 @@ def test_sample_with_overrides():
...
+        overrides=[
+            {"b": 4, "second": [{"c": 3}, {"c": 4}]},
+            {"b": 8, "second": [{"c": 6}]},
+        ]
+    )
+    assert collection.first.collect()["a"].to_list() == [0, 1]
+    assert collection.first.collect()["b"].to_list() == [4, 8]
+
+    assert collection.second is not None
+    assert collection.second.collect()["a"].to_list() == [0, 0, 1]
```
AndreasAlbertQC, line 131:
```suggestion
    assert collection.second.collect()["b"].to_list() == [4, 4, 8]
```
?

AndreasAlbertQC (approved):

tests/collection/test_sample.py
borchero: Ha, it should _not_ be `[4, 4, 8]` because the second member is not inlined 😄 The `b` in the overrides only refers to the first member.

borchero: I added the inverse test... it fails 1 out of 2**192 times but I guess that's fine 😂

PR #7, opened 2025-04-19
style: Enforce type annotations

borchero:
# Motivation

I think that it's generally good practice to enforce type annotations. This PR largely adds explicit `-> None` return annotations.

# Changes

- Set `disallow_untyped_defs = true` in the mypy options

### DIFF
--- a/docs/sites/quickstart.rst
+++ b/docs/sites/quickstart.rst
@@ -197,7 +197,7 @@ In this case, ``good`` remains to be a ``dy.DataFrame[HouseSchema]``, albeit wit
 The ``failure`` object is of type :class:`~dataframely.FailureInfo` and provides means to inspect
 the reasons for validation failures for invalid rows.
 
-Given the example data above and the schema that we defined, we know that rows 2, 3, 4, and 5 are invalid (0-indexed) -> None:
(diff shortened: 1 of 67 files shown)


docs/sites/quickstart.rst
```diff
@@ -197,7 +197,7 @@ In this case, ``good`` remains to be a ``dy.DataFrame[HouseSchema]``, albeit wit
 The ``failure`` object is of type :class:`~dataframely.FailureInfo` and provides means to inspect
 the reasons for validation failures for invalid rows.
 
-Given the example data above and the schema that we defined, we know that rows 2, 3, 4, and 5 are invalid (0-indexed):
+Given the example data above and the schema that we defined, we know that rows 2, 3, 4, and 5 are invalid (0-indexed) -> None:
```
delsner, line 200:
```suggestion
Given the example data above and the schema that we defined, we know that rows 2, 3, 4, and 5 are invalid (0-indexed):
```

delsner (approved):
Thanks!

docs/sites/quickstart.rst
borchero: Ha, guess my regex wasn't good enough 😂

PR #12, opened 2025-04-23
feat: Add `dy.Object` column type

### LINKED ISSUE #11: Add dy.Object for pl.Object column
Thank you for this lib, I am really eager to try it, from reading the doc, it looks really promising!

I notice the lake of `dy.Object` which block me from moving from my own tool: https://github.com/gab23r/polaroids

Would you be ok to add this column type ?

delsner:
# Motivation

Closes #11.

# Changes

Add `dy.Object` column type as requested by the community.
For now, we don't implement sampling which is not straightforward for arbitrary python objects.

### DIFF
--- a/dataframely/columns/object.py
+++ b/dataframely/columns/object.py
@@ -59,4 +59,6 @@ class Object(Column):
         raise NotImplementedError("PyArrow column cannot have 'Object' type.")
 
     def _sample_unchecked(self, generator: Generator, n: int) -> pl.Series:
-        return pl.Series([object() for _ in range(n)])
+        raise NotImplementedError(
+            "Random data sampling not implemented for 'Object' type."
+        )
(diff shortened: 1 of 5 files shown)


dataframely/columns/object.py
```diff
@@ -0,0 +1,63 @@
...
+        raise NotImplementedError("SQL column cannot have 'Object' type.")
+
+    @property
+    def pyarrow_dtype(self) -> pa.DataType:
+        raise NotImplementedError("PyArrow column cannot have 'Object' type.")
+
+    def _sample_unchecked(self, generator: Generator, n: int) -> pl.Series:
+        raise NotImplementedError(
+            "Random data sampling not implemented for 'Object' type."
+        )
```
borchero, line 63: We _could_ think about making this series all NULL? 🧐

borchero (approved):
Nice! 🥳

dataframely/columns/object.py
delsner: I'm not sure if this is a good default behaviour for sampling; it could cause confusion if the object's type is not nullable. Let's maybe wait for more feedback from the community before we implement sampling for `dy.Object`.

PR #14, opened 2025-04-24
feat: Allow nullable primary keys

### LINKED ISSUE #13: nullable=True is ignored if primary_key is True
The following snippet fails, where it should pass.

```py
class A(dy.Schema):
    a = dy.String(primary_key=True, nullable=True)

A.validate(pl.DataFrame({"a" : [None]}, schema={"a": pl.String}))
```

# RuleValidationError: 1 rules failed validation:
#  * Column 'a' failed validation for 1 rules:
#    - 'nullability' failed for 1 rows

gab23r:
fixes #13
# Motivation

Allow the user to define a nullable primary key column.

# Changes

`nullable` will default to `not primary_key` is not provided by the user, otherwise we keep user choice.

### DIFF
--- a/dataframely/columns/_base.py
+++ b/dataframely/columns/_base.py
@@ -28,7 +28,7 @@ class Column(ABC):
     def __init__(
         self,
         *,
-        nullable: bool = True,
+        nullable: bool | None = None,
         primary_key: bool = False,
         check: Callable[[pl.Expr], pl.Expr] | None = None,
         alias: str | None = None,
@@ -37,8 +37,10 @@ class Column(ABC):
         """
         Args:
             nullable: Whether this column may contain null values.
+                If `None`, the default behavior is as follows:
+                - If `primary_key` is `True`, `nullable` defaults to `False`.
+                - If `primary_key` is `False`, `nullable` defaults to `True`.
             primary_key: Whether this column is part of the primary key of the schema.
-                If ``True``, ``nullable`` is automatically set to ``False``.
             check: A custom check to run for this column. Must return a non-aggregated
                 boolean expression.
             alias: An overwrite for this column's name which allows for using a column
@@ -48,7 +50,7 @@ class Column(ABC):
                 internally sets the alias to the column's name in the parent schema.
             metadata: A dictionary of metadata to attach to the column.
         """
-        self.nullable = nullable and not primary_key
+        self.nullable = nullable if nullable is not None else not primary_key
         self.primary_key = primary_key
         self.check = check
         self.alias = alias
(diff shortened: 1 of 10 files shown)


borchero (commented):
Thanks for the immediate contribution following your issue @gab23r, great to see that you're keen on making dataframely work for your use case 😁

I thought about the fact that we're not allowing NULLs in primary keys again and I wanted to ask: are you actually interested in allowing NULL value for primary keys or do you use `primary_key=True` as a "workaround" for `unique=True`?

In my mind, this is slightly different. Polars has no concept of a "primary key" so it makes sense that it doesn't treat NULL values specially e.g. when using `n_unique()` or `group_by()`. However, a NULL value generally expresses "I don't know the value". Hence, it _does_ feel odd to allow NULL values in a primary key as an "unknown value" is typically not a good identifier for a row (after all, `NULL == NULL` also evaluates to `NULL`) 🤔 happy to hear your thoughts @AndreasAlbertQC @delsner

AndreasAlbertQC:
> Unique=True wouldn't make the job. As my table has multiple primary keys ( the column can have multiple nulls )

I think one should be able to implement this pretty straightforwardly by using a composite rule that enforces the composite uniqueness across multiple keys. I feel like this solution might be preferable because it avoids some of the murkiness discussed above:

```python 
class MySchema(dy.Schema):
    x = dy.Integer(nullable=True) # part of composite unique key
    y = dy.Integer(nullable=True)  # part of composite unique key
    z = dy.Integer(nullable=False)

    @dy.rule(group_by=["x","y"])
    def composite_unique():
        return pl.len() == 1
```

While testing this, though, I just noticed that in the special case of null x/y values, this does not appear to work as expected, which I think is a bug:

```python
# Should not pass validation
df = pl.DataFrame(
        {
            "x":[None,None],
            "y":[None,None],
            "z":[5,6],
        }
    )

# Passes validation!
MySchema.validate(df,cast=True)
```

Tracking this separately as #15

gab23r:
That would work indeed. I have no strong feelings about it even if I don't understand why we couldn't allow the user to have nullable primary keys if he explicitly asks for it. Are there technical things blocking for that ?

borchero:
@gab23r we have released #16 in v1.2.1 which allows for the custom workaround proposed by @AndreasAlbertQC. I think the simple fact that joins do not, by default, consider `null` values to be equal is a strong argument that we should not allow for nullable primary keys. Plenty of the functionality around collections would break if we lifted this requirement. I hope the custom workaround is sufficient for your needs.

That being said, I think we should still fail hard if the user provides both `primary_key=True` and `nullable=True` instead of silently ignoring it. Would you be interested in making another PR for that? :)

gab23r:
I understand. Raising an error would be better indeed. I can do a PR.

PR #16, opened 2025-04-25
fix: Make group rules work on nullable grouping columns

### LINKED ISSUE #15: @dy.rule with nullable columns in group_by does not work
When using nullable columns in grouped rules, the validation does not behave as expected.

Example:

```python 
class MySchema(dy.Schema):
    x = dy.Integer(nullable=True)

    @dy.rule(group_by=["x"])
    def unique():
        return pl.len() == 1

# Should not pass validation
df = pl.DataFrame(
        {
            "x":[1, None,None],
        }
    )

# Passes validation
MySchema.validate(df,cast=True)
```

I would have expected validation to fail because two rows have the same values of `x`. I have not debugged this yet, but suspect we are using a join somewhere after calling the `df.group_by(["x"]).agg(pl.len()==1)`, which then fails because of the nulls. If I use non-null duplicate values in the example, validation fails as expected.

AndreasAlbertQC:
# Motivation

Fixes #15 

# Changes

* Added `nulls_equal=True` to polars join used to accumulate group rule evaluation results.

### DIFF
--- a/dataframely/_rule.py
+++ b/dataframely/_rule.py
@@ -126,5 +126,7 @@ def _with_group_rules(lf: pl.LazyFrame, rules: dict[str, GroupRule]) -> pl.LazyF
     #    preserves the order of the left data frame.
     result = lf
     for group_columns, frame in group_evaluations.items():
-        result = result.join(frame, on=list(group_columns), how="left", nulls_equal=True)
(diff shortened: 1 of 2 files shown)


borchero (approved):
Thanks for fixing!

PR #18, opened 2025-04-25
fix: Make enums tolerate multiple sequence types

### LINKED ISSUE #17: Roundtripping enums through parquet fails unexpectedly if the enum is defined via a tuple
This one's pretty weird:

```python
from dataframely import Schema, Enum
import polars as pl

# Schema with enum, note that vals is a tuple!
vals = ("a","b")
class S(Schema):
    x = Enum(vals)

# All good here
df = pl.DataFrame(
    {"x":["a","b"]}
)
validated = S.validate(df,cast=True)


# Now roundtrip through parquet
fpath = "/tmp/data.parquet"
validated.write_parquet(fpath)
readback = pl.read_parquet(fpath)

# This seems to still be OK
assert validated.dtypes==readback.dtypes

# This now fails ?!
S.validate(readback)

# dataframely.exc.DtypeValidationError: 1 columns have an invalid dtype:
#  - 'x': got dtype 'Enum(categories=['a', 'b'])' but expected 'Enum(categories=['a', 'b'])'
```

If I make `vals` a `list` instead of a `tuple`, everything works fine.

AndreasAlbertQC:
# Motivation

Fixes #17, which actually has nothing to do with parquet. The parquet connection just happened because the enum categories are always a `list` after going through parquet.

# Changes

* Added `list` cast to ensure that `Enum.categories` is always stored as a list internally.
* Added a test that fails without the fix. @delsner @borchero : I wasn't 100% sure what file the test should be in. Let me know if you think it should go somewhere else.

### DIFF
--- a/dataframely/columns/enum.py
+++ b/dataframely/columns/enum.py
@@ -49,7 +49,7 @@ class Enum(Column):
             alias=alias,
             metadata=metadata,
         )
-        self.categories = categories
+        self.categories = list(categories)
 
     @property
     def dtype(self) -> pl.DataType:
--- a/tests/column_types/test_enum.py
+++ b/tests/column_types/test_enum.py
@@ -52,3 +52,12 @@ def test_valid_cast(
     schema = create_schema("test", {"a": enum})
     df = df_type(data)
     assert schema.is_valid(df, cast=True) == valid
+
+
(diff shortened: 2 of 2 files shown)


borchero (approved):
Ha, weird edge case 👀 thanks for fixing!

PR #19, opened 2025-04-26
feat: Allow for multiple column level checks with custom naming

heshamdar:
# Motivation

Currently only a single column specific check is allowed. Although multiple checks could be bundled into one function this means that the specific failing rule is opaque.

Additionally the column specific checks are given the default name `check` so is not immediately apparent what specifically fails without referring to the implementation.

# Changes

- Added support for accepting a list or dictionary of checks for the column `check` argument
- In the case of a dictionary, the string is given as the rule name
- In the case of a single `callable` or list of `callable`s the `__name__` attribute is used, so can provide an understandable name if a named function is defined
    - where a lambda or list of lambda functions are used, the name is given as `check` or `check_{n}` where `n` is the index

The added tests show examples of these

# Note
- Although this works, I haven't checked all related functionality so there might be downstream issues


dataframely/columns/_base.py
```diff
@@ -30,7 +30,10 @@ def __init__(
         *,
         nullable: bool = True,
         primary_key: bool = False,
-        check: Callable[[pl.Expr], pl.Expr] | None = None,
+        check: Callable[[pl.Expr], pl.Expr]
+        | list[Callable[[pl.Expr], pl.Expr]]
+        | dict[str, Callable[[pl.Expr], pl.Expr]]
+        | None = None,
```
borchero, line 36: [nit] layout

```suggestion
        check: (
            Callable[[pl.Expr], pl.Expr]
            | list[Callable[[pl.Expr], pl.Expr]]
            | dict[str, Callable[[pl.Expr], pl.Expr]]
            | None
        ) = None,
```

dataframely/columns/_base.py
```diff
@@ -97,7 +100,25 @@ def validation_rules(self, expr: pl.Expr) -> dict[str, pl.Expr]:
...
-            result["check"] = self.check(expr)
+            if isinstance(self.check, dict):
+                for k, v in self.check.items():
+                    result[k] = v(expr)
+
+            elif isinstance(self.check, list):
+                for num, i in enumerate(self.check):
+                    callable_name = i.__name__
+                    if callable_name == "<lambda>":
+                        callable_name = f"check__{num}"
```
borchero, line 111: What's the rationale for using two underscores here?

dataframely/columns/_base.py
```diff
@@ -97,7 +100,25 @@ def validation_rules(self, expr: pl.Expr) -> dict[str, pl.Expr]:
...
+                    result[callable_name] = i(expr)
+
+            elif callable(self.check):
+                callable_name = self.check.__name__
```
borchero, line 115: I think it's a reasonable idea to try and use the callable's `__name__` but would be interested if @delsner and/or @AndreasAlbertQC agree :smile:

dataframely/columns/_base.py
```diff
@@ -97,7 +100,25 @@ def validation_rules(self, expr: pl.Expr) -> dict[str, pl.Expr]:
...
+                    result[k] = v(expr)
+
+            elif isinstance(self.check, list):
+                for num, i in enumerate(self.check):
+                    callable_name = i.__name__
+                    if callable_name == "<lambda>":
+                        callable_name = f"check__{num}"
+                    result[callable_name] = i(expr)
+
+            elif callable(self.check):
```
borchero, line 114: Given that we expect the user to leverage `mypy`, we could go into the `else` block here. Otherwise, we'd need to add an `else` block raising an exception (and exclude that from coverage which isn't great IMO).

dataframely/columns/_base.py
borchero: (applies to all subclasses as well)

tests/columns/test_check.py
```diff
@@ -8,12 +8,54 @@
 
 
 class CheckSchema(dy.Schema):
-    a = dy.Int64(check=lambda col: (col < 5) | (col > 10))
-    b = dy.String(min_length=3, check=lambda col: col.str.contains("x"))
+    a = dy.Int64(check={"check": lambda col: (col < 5) | (col > 10)})
+    b = dy.String(min_length=3, check={"check": lambda col: col.str.contains("x")})
```
borchero, line 12: Is there a need to change this? The change should have no effect, right?

tests/columns/test_check.py
```diff
@@ -8,12 +8,54 @@
...
+
+
+def test_check_names() -> None:
+    def str_starts_with_a(col: pl.Expr) -> pl.Expr:
+        return col.str.starts_with("a")
+
+    def str_end_with_z(col: pl.Expr) -> pl.Expr:
+        return col.str.ends_with("z")
+
+    class MultiCheckSchema(dy.Schema):
```
borchero, line 29: Can you add a column where we pass checks as a list of lambdas? That'll maintain 100% code coverage (see [here](https://app.codecov.io/gh/Quantco/dataframely/pull/19?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Quantco#c1e424ec368c65c706a8a51dcc3fa5b3-R111))

borchero (commented):
Thanks a lot for your contribution @heshamdar -- great to see that you seem to have quickly found your way around the codebase all by yourself! :)

The change makes a lot of sense to me, I already had the same thought at some point. I just left some small comments :)

dataframely/columns/_base.py
heshamdar: No strong reason really except spaces typically use a single underscore so to separate in the off-chance that a named function ends with an integer. Happy to change this to something else!

tests/columns/test_check.py
heshamdar: Correct, originally I didn't handle the nameless case so they ended up with `<lambda>` as the name, so changed the test. Now that it sets nameless funcs to `check` I'll revert these tests and they should pass

dataframely/columns/_base.py
AndreasAlbertQC: I mostly think we need to handle duplicate names for every branch the same as we currently do for `callable_name`. 

I can think of two ways this would happen:
1. Imports with rename:
```python
from x import check as check1
from y import check as check2
# will be true
assert x.__name__ == y.__name__
```
2. decorators:
```python
def mydec(fun):
  def wrap(fun):
     ...
  return wrap

@mydec
def fun1():
  pass

@mydec
def fun2():
  pass

assert fun1.__name__ == 'wrap'
assert fun2.__name__ == 'wrap'
```

How about we:
1. check if `__name__` is duplicated
2. if it is, we postfix all occurrences  of that `__name__` with a counter as done for lambdas. IMO the cleanest would be if the counter started from 0 for every duplicate name so I get `["x","y_0","y_1"]` instead of `["x","y_1",...]`.

dataframely/columns/_base.py
```diff
@@ -97,7 +102,25 @@ def validation_rules(self, expr: pl.Expr) -> dict[str, pl.Expr]:
         if not self.nullable:
             result["nullability"] = expr.is_not_null()
         if self.check is not None:
-            result["check"] = self.check(expr)
+            if isinstance(self.check, dict):
+                for k, v in self.check.items():
+                    result[k] = v(expr)
+
+            elif isinstance(self.check, list):
+                for num, i in enumerate(self.check):
```
AndreasAlbertQC, line 110:
```suggestion
                for i, callable in enumerate(self.check):
```
I'd find this more intuitive

dataframely/columns/_base.py
AndreasAlbertQC: > in the off-chance that a named function ends with an integer.

Smart 👍

dataframely/columns/_base.py
borchero: (2) should not usually be an issue if decorators correctly use `functools.wraps` but this is besides the point, I think :smile:

I like the idea to postfix on duplicates :)

dataframely/columns/_base.py
borchero: I don't think I understand it yet, could one of you provide an example where this fixes an issue? 🙈

dataframely/columns/_base.py
```diff
@@ -30,7 +30,12 @@ def __init__(
         *,
         nullable: bool = True,
         primary_key: bool = False,
-        check: Callable[[pl.Expr], pl.Expr] | None = None,
+        check: (
```
borchero, line 33: We'll still need to adjust the docstring for this parameter.

dataframely/columns/_base.py
```diff
@@ -97,7 +102,25 @@ def validation_rules(self, expr: pl.Expr) -> dict[str, pl.Expr]:
         if not self.nullable:
             result["nullability"] = expr.is_not_null()
         if self.check is not None:
-            result["check"] = self.check(expr)
+            if isinstance(self.check, dict):
+                for k, v in self.check.items():
+                    result[k] = v(expr)
```
borchero, line 107: And I was wondering whether we should prefix every check with `check_` to avoid name collisions.

dataframely/columns/_base.py
heshamdar: If someone has a function that checks if a value is greater than 0, they might call it `greater_than_0`. 

With a single underscore for indicating index we would get `check_0`, so maybe someone looking at the end report might get confused and think it's checking for the number 0. 

With a double underscore it would (hopefully) implicitly suggest a different meaning.

This does make me think forcing the user to define a name in all cases would actually be clearer, albeit increasing their workload.

dataframely/columns/_base.py
```diff
@@ -79,6 +93,45 @@ def validate_dtype(self, dtype: PolarsDataType) -> bool:
...
+        Returns:
+            List of unique names corresponding to the rule callables.
+        """
+        # Extract base names
+        base_rule_names = []
+        for rule in rules:
+            base_rule_name = rule.__name__
+            if base_rule_name == "<lambda>":
+                base_rule_name = "check"
+            base_rule_names.append(base_rule_name)
```
AndreasAlbertQC, line 117:
```suggestion
        # Extract base names
        base_names = [
            rule.__name__
            if rule.__name__ != "<lambda>"
            else "check"
            for rule in rules
        ]
```

dataframely/columns/_base.py
```diff
@@ -79,6 +93,45 @@ def validate_dtype(self, dtype: PolarsDataType) -> bool:
...
+
+        # Count occurrences using Counter
+        name_counts = Counter(base_rule_names)
```
AndreasAlbertQC, line 120:
```suggestion
        name_counts = Counter(base_names)
```

dataframely/columns/_base.py
```diff
@@ -79,6 +93,45 @@ def validate_dtype(self, dtype: PolarsDataType) -> bool:
...
+        dynamic_name_counter: dict[str, int] = {}
+        for base_rule_name in base_rule_names:
+            if name_counts[base_rule_name] > 1:
+                idx = dynamic_name_counter.get(base_rule_name, 0)
+                rule_names_with_counts.append(f"{base_rule_name}{count_separator}{idx}")
+                dynamic_name_counter[base_rule_name] = idx + 1
+            else:
+                rule_names_with_counts.append(base_rule_name)
+
+        return rule_names_with_counts
```
AndreasAlbertQC, line 133:
```suggestion
        # Append suffixes to names that are duplicated
        final_names = []
        duplicate_counter: dict[str, int] = {}
        for name in base_names:
            if name_counts[name] > 1:
                postfix = duplicate_counter.get(name, 0)
                final_names.append(f"{name}{count_separator}{postfix}")
                duplicate_counter[name]+=1
           else:
                final_names.append(name)

        return final_names
```

dataframely/columns/_base.py
```diff
@@ -79,6 +93,45 @@ def validate_dtype(self, dtype: PolarsDataType) -> bool:
 
     # ---------------------------------- VALIDATION ---------------------------------- #
 
+    def _get_rule_names(
+        self, rules: list[Callable[[pl.Expr], pl.Expr]], count_separator: str = "__"
```
AndreasAlbertQC, line 97:
```suggestion
        self, rules: list[Callable[[pl.Expr], pl.Expr]]
```
IMO it would be OK to just commit to this being one value, since I don't think we'll every use different ones.
Should we just make this a constant `RULE_NAME_SUFFIX_SEPARATOR`?

dataframely/columns/_base.py
```diff
@@ -97,7 +150,28 @@ def validation_rules(self, expr: pl.Expr) -> dict[str, pl.Expr]:
...
         if self.check is not None:
-            result["check"] = self.check(expr)
+            if isinstance(self.check, dict):
+                for rule_name, rule_callable in self.check.items():
+                    result[rule_name] = rule_callable(expr)
+
+            elif isinstance(self.check, list):
+                # Get unique names for rules from callables
+                rule_names = self._get_rule_names(
+                    rules=self.check, count_separator="__"
```
AndreasAlbertQC, line 160:
```suggestion
                    rules=self.check
```

AndreasAlbertQC (commented):
Thanks for your nice work @heshamdar! I like the new logic. I'm just leaving a few suggestions to help further increasing readability.

AndreasAlbertQC (approved):
Thanks for incorporating the feedback @heshamdar ! LGTM.

dataframely/columns/datetime.py
```diff
@@ -149,7 +162,10 @@ def __init__(
         max: dt.time | None = None,
         max_exclusive: dt.time | None = None,
         resolution: str | None = None,
-        check: Callable[[pl.Expr], pl.Expr] | None = None,
+        check: Callable[[pl.Expr], pl.Expr]
+        | list[Callable[[pl.Expr], pl.Expr]]
+        | dict[str, Callable[[pl.Expr], pl.Expr]]
+        | None = None,
```
borchero, line 168:
```suggestion
        check: (
            Callable[[pl.Expr], pl.Expr]
            | list[Callable[[pl.Expr], pl.Expr]]
            | dict[str, Callable[[pl.Expr], pl.Expr]]
            | None
        ) = None,
```

dataframely/columns/datetime.py
```diff
@@ -265,7 +281,10 @@ def __init__(
         max: dt.datetime | None = None,
         max_exclusive: dt.datetime | None = None,
         resolution: str | None = None,
-        check: Callable[[pl.Expr], pl.Expr] | None = None,
+        check: Callable[[pl.Expr], pl.Expr]
+        | list[Callable[[pl.Expr], pl.Expr]]
+        | dict[str, Callable[[pl.Expr], pl.Expr]]
+        | None = None,
```
borchero, line 287:
```suggestion
        check: (
            Callable[[pl.Expr], pl.Expr]
            | list[Callable[[pl.Expr], pl.Expr]]
            | dict[str, Callable[[pl.Expr], pl.Expr]]
            | None
        ) = None,
```

dataframely/columns/datetime.py
```diff
@@ -371,7 +390,10 @@ def __init__(
         max: dt.timedelta | None = None,
         max_exclusive: dt.timedelta | None = None,
         resolution: str | None = None,
-        check: Callable[[pl.Expr], pl.Expr] | None = None,
+        check: Callable[[pl.Expr], pl.Expr]
+        | list[Callable[[pl.Expr], pl.Expr]]
+        | dict[str, Callable[[pl.Expr], pl.Expr]]
+        | None = None,
```
borchero, line 396:
```suggestion
        check: (
            Callable[[pl.Expr], pl.Expr]
            | list[Callable[[pl.Expr], pl.Expr]]
            | dict[str, Callable[[pl.Expr], pl.Expr]]
            | None
        ) = None,
```

dataframely/columns/datetime.py
borchero: Could you also adjust the docstring here?

dataframely/columns/_base.py
```diff
@@ -79,6 +95,42 @@ def validate_dtype(self, dtype: PolarsDataType) -> bool:
 
     # ---------------------------------- VALIDATION ---------------------------------- #
 
+    def _get_rule_names(self, rules: list[Callable[[pl.Expr], pl.Expr]]) -> list[str]:
```
borchero, line 98: This is really only about _check_ rules. Could we name this accordingly, e.g.

```suggestion
    def _derive_check_rule_names(self, rules: list[Callable[[pl.Expr], pl.Expr]]) -> list[str]:
```

borchero: And [nit], can we move this method below the `validation_rules` method (within the `VALIDATION` section)? I find it easier to read through code if the most important stuff comes first :smile:

dataframely/columns/_base.py
```diff
@@ -97,7 +149,26 @@ def validation_rules(self, expr: pl.Expr) -> dict[str, pl.Expr]:
...
+
+                # Apply each rule with its corresponding name
+                for rule_name, rule_callable in zip(rule_names, self.check):
+                    result[rule_name] = rule_callable(expr)
+
+            else:
+                # Get rule name, default to "check" for lambdas
+                rule_name = self.check.__name__
+                if rule_name == "<lambda>":
+                    rule_name = "check"
```
borchero, line 168: Can we dispatch to our private method here to not have the `"check"` fallback implemented twice?

dataframely/columns/_base.py
```diff
@@ -97,7 +149,26 @@ def validation_rules(self, expr: pl.Expr) -> dict[str, pl.Expr]:
...
+
+            elif isinstance(self.check, list):
+                # Get unique names for rules from callables
+                rule_names = self._get_rule_names(rules=self.check)
+
+                # Apply each rule with its corresponding name
+                for rule_name, rule_callable in zip(rule_names, self.check):
+                    result[rule_name] = rule_callable(expr)
+
+            else:
```
borchero, line 164: [nit] since this is a nested conditional, I rather find that the spacing between the cases makes the code harder to read. Let's remove it

```suggestion
            elif isinstance(self.check, list):
                # Get unique names for rules from callables
                rule_names = self._get_rule_names(rules=self.check)

                # Apply each rule with its corresponding name
                for rule_name, rule_callable in zip(rule_names, self.check):
                    result[rule_name] = rule_callable(expr)
            else:
```

borchero (commented):
Thank you for the updates, this looks great! :) I just left a few small comments after which I'm confident we can merge 🚀

dataframely/columns/_base.py
```diff
@@ -97,9 +113,63 @@ def validation_rules(self, expr: pl.Expr) -> dict[str, pl.Expr]:
         if not self.nullable:
             result["nullability"] = expr.is_not_null()
         if self.check is not None:
-            result["check"] = self.check(expr)
+            if isinstance(self.check, dict):
+                for rule_name, rule_callable in self.check.items():
+                    result[rule_name] = rule_callable(expr)
+            else:
+                # wrap single callable in a list
+                if isinstance(self.check, list):
```
heshamdar, line 121: not sure if there's a more elegant way to handle this, the type checker doesn't seem to like 

```
list_of_rules = self.check
if not isinstance(list_of_rules, list):
    list_of_rules = [self.check]
```

borchero: I think that's fine

borchero: I made a tiny adjustment by using a ternary operator

dataframely/columns/_base.py
borchero: I still want to have the `check_` prefix to ensure that the user cannot provide a name that is already used by other validation rules

borchero (commented):
Just one more comment, see the last open conversation above :)

borchero (approved):
Thanks for your work on this @heshamdar, sorry this took a bit! I still adjusted the names of the checks a little bit to always prepend `check__` (note the double underscore to align with the suffix) even if the method name starts with `check`. I think this will be very useful now :)

heshamdar:
> Thanks for your work on this @heshamdar, sorry this took a bit! I still adjusted the names of the checks a little bit to always prepend `check__` (note the double underscore to align with the suffix) even if the method name starts with `check`. I think this will be very useful now :)

Great, happy to contribute! 🙂

PR #24, opened 2025-04-30
fix: Generate nulls for Nullable list/struct in `sample`

gab23r:
# Motivation

`sample` should reflects the nullable property of listcolumns.

# Changes

Add `_apply_null_mask` in ``List._sample_unchecked`.

series.scatter is not possible on a list series
```py
pl.Series(dtype=pl.List(pl.Int64)).scatter([0], None) 
# ComputeError: not yet implemented for dtype: list[i64]
```
So I used a pl.when().then() approach which ends up being much faster!

```py

import numpy as np
import polars as pl
s = pl.Series(range(10_000_000))
null_probability = 0.1

numpy_generator = np.random.default_rng()
random_values = numpy_generator.random(s.len())


%%timeit
null_mask = random_values < null_probability
s.scatter(np.where(null_mask)[0], None)
# 35.2 ms ± 3.66 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit
null_mask = pl.Series(random_values) > null_probability
pl.select(pl.when(null_mask).then(s)).to_series()
# 6.04 ms ± 190 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

```
EDIT: add the same feature for struct

### DIFF
--- a/dataframely/columns/list.py
+++ b/dataframely/columns/list.py
@@ -147,4 +147,7 @@ class List(Column):
                 chain([0], element_lengths.cum_sum()), element_lengths
             )
         ]
-        return pl.Series(list_elements)
+        # Finally, apply a null mask
+        return generator._apply_null_mask(
+            pl.Series(list_elements), null_probability=self._null_probability
+        )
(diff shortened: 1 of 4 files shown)


borchero (approved):
Nice, thank you @gab23r!

PR #27, opened 2025-05-01
feat: Add `dy.Array` column type

jjurm:
# Motivation

Dataframely currently doesn't support the `pl.Array` column type, and so dataframes using array columns cannot be validated. I'd like to be able to validate those, too.

# Changes

This PR implements the `dy.Array` column type corresponding to polars' `pl.Array` type.

The implementation is very similar to the `dy.List`, however, polars List differs from Array mostly in the following ways:
- Lists are flat, whereas Arrays can have have shapes with multiple axes
- Arrays must have fixed size (or shape) whereas List elements can vary in length
- `pl.Expr.arr` has less functionality compared to `pl.Expr.list`, most notably there is no `.arr.eval()` function, which complicates validating rules. One option is to recursively convert the array column to nested lists just for validation, but since the performance would be suboptimal, I left this part unimplemented and only allowed Array's inner column types without any validation rules.

Would be happy to hear any feedback!


dataframely/columns/array.py
```diff
@@ -0,0 +1,128 @@
...
+            alias: An overwrite for this column's name which allows for using a column
+                name that is not a valid Python identifier. Especially note that setting
+                this option does _not_ allow to refer to the column with two different
+                names, the specified alias is the only valid name.
+            metadata: A dictionary of metadata to attach to the column.
+        """
+        if primary_key:
+            raise ValueError(
+                "`primary_key=True` is not yet supported for the Array type."
+            )
```
borchero, line 52: I don't think we need to cover this case specifically if the type annotation already prevents passing in `True`.

dataframely/columns/array.py
```diff
@@ -0,0 +1,128 @@
...
+        ):
+            raise ValueError(
+                "`primary_key=True` is not yet supported for inner types of the Array type."
+            )
+
+        # We disallow validation rules on the inner type since Polars arrays currently don't support .eval(). Converting
+        # to a list and calling .list.eval() is possible, however, since the shape can have multiple axes, the recursive
+        # conversion could have significant performance impact. Hence, we simply disallow inner validation rules.
+        # Another option would be to allow validation rules only for sampling, but not enforce them.
+        if inner.validation_rules(pl.col(alias or "")):
```
borchero, line 66: [nit] I'd probably use a literal as "stub" here :smile:

```suggestion
        if inner.validation_rules(pl.lit(None)):
```

dataframely/columns/array.py
```diff
@@ -0,0 +1,128 @@
...
+
+        super().__init__(
+            nullable=nullable,
+            primary_key=False,
+            check=check,
+            alias=alias,
+            metadata=metadata,
+        )
+        self.inner = inner
+        self.shape = self._ensure_tuple(shape)
```
borchero, line 79: [nit] No need for the indirection via the method IMO

```suggestion
        self.shape = shape if isinstance(shape, tuple) else (shape,)
```

dataframely/columns/array.py
```diff
@@ -0,0 +1,128 @@
...
+        return pl.Array(self.inner.dtype, self.shape)
+
+    @staticmethod
+    def _ensure_tuple(value: int | tuple[int, ...]) -> tuple[int, ...]:
+        if isinstance(value, tuple):
+            return value
+        return (value,)
+
+    def sqlalchemy_dtype(self, dialect: sa.Dialect) -> sa_TypeEngine:
+        # NOTE: We might want to add support for PostgreSQL's ARRAY type or use JSON in the future.
```
borchero, line 92: 👍🏼 I think we're generally still lacking support for more advanced dtypes across various RDBMS 😅

dataframely/columns/array.py
```diff
@@ -0,0 +1,128 @@
...
+        raise NotImplementedError("SQL column cannot have 'Array' type.")
+
+    def _pyarrow_dtype_of_shape(self, shape: Sequence[int]) -> pa.DataType:
+        if shape:
+            size, *rest = shape
+            return pa.list_(self._pyarrow_dtype_of_shape(rest), size)
+        else:
+            return self.inner.pyarrow_dtype
```
borchero, line 100: I think we have to be more explicit about the check here as, otherwise, we'd end up in the `else` branch when the user provides an "empty shape" (i.e. `shape=0`). Could you add a test that we correctly return a list if we pass `shape=0`?

```suggestion
        if len(shape) > 0:
            size, *rest = shape
            return pa.list_(self._pyarrow_dtype_of_shape(rest), size)
        else:
            return self.inner.pyarrow_dtype
```

dataframely/columns/array.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @property
+    def pyarrow_dtype(self) -> pa.DataType:
+        return self._pyarrow_dtype_of_shape(self.shape)
+
+    def _sample_unchecked(self, generator: Generator, n: int) -> pl.Series:
+        # Sample the inner elements in a flat series
+        n_elements = n * cast(int, pl.Series(self.shape).product())
```
borchero, line 108: Can we use `math.prod` here?

```suggestion
        n_elements = n * math.prod(self.shape)
```

dataframely/columns/array.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # Turn the "long" series into a series with arrays
+        it = iter(all_elements)
+
+        def _reshape_recursive(shape: Sequence[int]) -> Any:
+            if not shape:
+                return next(it)
+            rest: list[int]
+            n, *rest = shape
+            return [_reshape_recursive(rest) for _ in range(n)]
```
borchero, line 119: I think this part here can be easier with native polars functionality (see https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.reshape.html) along the lines of:

```python
all_elements.reshape((n,) + self.shape)

File: dataframely/testing/const.py [hunk 8]
```diff
@@ -47,3 +47,7 @@
 ALL_COLUMN_TYPES: list[type[dc.Column]] = (
     [dc.Any] + COLUMN_TYPES + SUPERTYPE_COLUMN_TYPES
 )
+
+NO_VALIDATION_COLUMN_TYPES: list[type[dc.Column]] = [
```
borchero, line 51: Is this supposed to be all column types which do not have a validation rule by default or is there another idea? Could you add a comment what this is and why it exists?

tests/column_types/test_array.py
```diff
@@ -0,0 +1,170 @@
...
+        pl.DataFrame(
+            {"a": [[[1]], [[2]], [[3]]]},
+            schema={
+                "a": pl.Array(pl.Int64, (1, 1)),
+            },
+        )
+    )
+
+
+def test_array_with_pk() -> None:
```
borchero, line 131: I think we can get rid of this test here as we already need to fight the type checker :smile:

dataframely/columns/array.py
```diff
@@ -0,0 +1,128 @@
...
+    """A fixed-shape array column."""
+
+    def __init__(
+        self,
+        inner: Column,
+        shape: int | tuple[int, ...],
+        *,
+        nullable: bool = True,
+        # polars doesn't yet support grouping by arrays,
+        # see https://github.com/pola-rs/polars/blob/810df456/crates/polars-core/src/frame/group_by/into_groups.rs#L365
```
borchero, line 28: Ideally, we'd be linking an issue here so that we can trace more easily when it might be resolved. If none exists, could you create one? :)

borchero (commented):
Thank you, this looks great already! I left a couple of small comments but I think it is almost ready :)

Great job finding all the right places to add code, esp. in the tests, and thanks for maintaining the 100% coverage 😁🚀

dataframely/columns/array.py
jjurm: my idea was that the following situation might be very common:
- a library user being used to the primary_key argument accepting any boolean value, and
- the same user not using typing checks
in those cases, it is easy to just write `primary_key=True`, the error would show up later and the message from polars would be more difficult to decipher. Hence the proactive exception.

Let me know if you still think this is unnecessary.

borchero: > the same user not using typing checks

So far, we've ignored this setting as our stance was that typing is so integral to the use of dataframely that it doesn't make sense to use it without. This choice is mirrored throughout the codebase (which has also prevented writing tests that require ignoring the type checker).

For consistency, I'd propose to get rid of it. If we ever want to make dataframely nicely usable without typing, we'd have to touch a couple places, I think.

delsner (changes requested):
Could you please also update the mypy plugin and `tests/test_typing.py` similar to #29? Thanks!

dataframely/columns/array.py
jjurm: The current code works because shape is a `tuple[int, ...]` and cannot be an `int`, and `bool()` of a non-emtpy tuple evaluates to `True`.

I parametrised https://github.com/Quantco/dataframely/blob/add-array-column/tests/columns/test_pyarrow.py#L53-L61 to demonstrate that this works on `shape=0`, too.

Let me know if you still prefer the explicit `len(shape) > 0`.

dataframely/columns/array.py
jjurm: I just opened https://github.com/pola-rs/polars/issues/22574 and it seems it will be implemented soon!

borchero: Damn, that was fast :smile

dataframely/columns/array.py
borchero: > bool() of a non-emtpy tuple evaluates to True.

Ah sorry, I somehow thought that `bool((0,))` evaluates to `False` but it doesn't :smile: no need to change then!

borchero (approved):
Looks good to me (modulo the mypy plugin that @delsner mentioned), thanks! 🚀

dataframely/columns/array.py
```diff
@@ -0,0 +1,105 @@
...
+    """A fixed-shape array column."""
+
+    def __init__(
+        self,
+        inner: Column,
+        shape: int | tuple[int, ...],
+        *,
+        nullable: bool = True,
+        # polars doesn't yet support grouping by arrays,
+        # see https://github.com/pola-rs/polars/issues/22574
```
delsner, line 29: nit: Should we maybe move this down to where we have to make the update once this is released?

```python
super().__init__(
    nullable=nullable,
    # polars doesn't yet support grouping by arrays,
    # xref: https://github.com/pola-rs/polars/issues/22574
    primary_key=False,
    check=check,
    alias=alias,
    metadata=metadata,
)
```

delsner (approved):
Great work, thanks! 🚀 I added a typing test and updated the mypy plugin code.

dataframely/columns/array.py
borchero: I think it's fine, we need to adjust both places anyways.

PR #29, opened 2025-05-02
fix: Add `dy.Object` to mypy plugin

delsner:
# Motivation

We forgot to add `dy.Object` to the mypy plugin in #11.

# Changes

Add `dy.Object` to mypy plugin.


borchero (approved):
Thanks! We'll probably need to think about the mypy plugin in #27 as well.

PR #34, opened 2025-05-09
fix: Correctly set dunder name for schemas with at least one rule

MoritzPotthoffQC:
# Motivation

For schemas that define at least one rule, `Schema.__name__` is not set correctly (see CI on the first commit).

# Changes

- Avoid shadowing `name` with rule names in the base schema constructor. 
- Add tests

### DIFF
--- a/dataframely/_base_schema.py
+++ b/dataframely/_base_schema.py
@@ -98,7 +98,7 @@ class SchemaMeta(ABCMeta):
             )
 
         # 2) Check that the columns referenced in the group rules exist.
-        for name, rule in result.rules.items():
+        for rule_name, rule in result.rules.items():
             if isinstance(rule, GroupRule):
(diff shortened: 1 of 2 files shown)


dataframely/_base_schema.py
```diff
@@ -116,16 +116,16 @@ def __new__(
         # empty data frame and checking for the evaluated dtypes.
         if len(result.rules) > 0:
             lf_empty = pl.LazyFrame(
-                schema={name: col.dtype for name, col in result.columns.items()}
```
MoritzPotthoffQC, line 119: This change is not necessary, but I still found it nice to be precise in the inner scope.

AndreasAlbertQC (approved):
Subtle! Thanks @MoritzPotthoffQC.

borchero:
Phew whoopsie 👀

PR #35, opened 2025-05-09
docs: Document our approach to breaking changes

AndreasAlbertQC:
# Motivation

We want to agree on how to deal with breaking changes.

# Changes

* Added a description of how we use `FutureWarnings` to communicate about incoming breaking changes.

### DIFF
--- a/docs/sites/breaking.rst
+++ b/docs/sites/breaking.rst
@@ -1,0 +1,10 @@
+Breaking changes
+================
+
+Breaking changes to existing APIs are naturally disruptive to users, but sometimes
+necessary in order to introduce new functionality.
+When we intend to make breaking changes to user-facing APIs in dataframely, we aim to make it easy for users to anticipate such changes and migrate their code to the new behavior.
+
+``FutureWarnings``
+----------------
+Wherever possible, we introduce `FutureWarnings <https://docs.python.org/3/library/exceptions.html#FutureWarning>`_ before the breaking changes take effect. Warnings are the most direct and effective tool at our disposal for reaching users directly. We therefore generally recommend that users do not silence such warnings explicitly. We believe that the best way of silencing a warning should be to migrate to new behavior early. However, we also understand that the need for migration may catch users at an inconvenient time, and a temporary band aid solution might be required. Users can disable ``FutureWarnings`` either through `python builtins <https://docs.python.org/3/library/warnings.html#warnings.filterwarnings>`_, builtins from tools `like pytest <https://docs.pytest.org/en/stable/how-to/capture-warnings.html#controlling-warnings>`_ , or by setting the ``DATAFRAMELY_NO_FUTURE_WARNINGS`` environment variable to ``true`` or ``1``.
--- a/docs/sites/development.rst
+++ b/docs/sites/development.rst
@@ -42,6 +42,7 @@ documentation and then open it in your web browser using the commands below:
 .. code-block:: bash
 
     # Run build
+    pixi run -e docs postinstall
     pixi run docs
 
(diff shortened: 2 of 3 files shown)


docs/sites/development.rst
```diff
@@ -42,6 +42,7 @@ documentation and then open it in your web browser using the commands below:
 .. code-block:: bash
 
     # Run build
+    pixi run -e docs postinstall
```
AndreasAlbertQC, line 45: just noticed this locally: need to postinstall first because the `docs` feature is not part of the default environment.

docs/sites/breaking.rst
```diff
@@ -0,0 +1,10 @@
+Breaking changes
+================
+
+Breaking changes to existing APIs are naturally disruptive to users, but sometimes
+necessary in order to introduce new functionality.
+When we intend to make breaking changes to user-facing APIs in dataframely, we aim to make it easy for users to anticipate such changes and migrate their code to the new behavior.
+
+``FutureWarnings``
+----------------
```
borchero, line 9: Does this render correctly?

```suggestion
``FutureWarnings``
------------------
```

docs/sites/breaking.rst
```diff
@@ -0,0 +1,10 @@
+Breaking changes
```
borchero, line 1: I'd like to talk about major releases here as well. We should either (1) commit to never _actually_ making breaking changes without releasing a new major version or, at least, (2) promise a warning for X minor releases/Y months before we make the breaking change. I'd personally propose to go for (1).

AndreasAlbertQC: (1) sounds great. For nullability, we will do (2) though, right? So I'm not sure I can put (1) in the text xD

docs/sites/breaking.rst
AndreasAlbertQC: I think thats more of a guideline that a rule, it seems to render correctly either way (on my machine :tm:), anyway accepted the suggestoin

docs/sites/breaking.rst
borchero: I think it would be fine to do a major release in a couple months 🤔 but happy to talk about this, we definitely shouldn't document something that we're going to deviate from though :smile:

docs/sites/versioning.rst
```diff
@@ -0,0 +1,20 @@
...
+We therefore recommend that users are particularly vigilant when updating their environments to a newer major release of `dataframely`.
+As always, automated testing is useful here, but we also recommend checking the release notes `published on GitHub <https://github.com/Quantco/dataframely/releases>`_.
+
+In order to give users a heads-up before breaking changes are released, we introduce `FutureWarnings <https://docs.python.org/3/library/exceptions.html#FutureWarning>`_ .
+Warnings are the most direct and effective tool at our disposal for reaching users directly.
+We therefore generally recommend that users do not silence such warnings explicitly, but instead migrate their code proactively, whenever possible.
+However, we also understand that the need for migration may catch users at an inconvenient time, and a temporary band aid solution might be required.
+Users can disable ``FutureWarnings`` either through `python builtins <https://docs.python.org/3/library/warnings.html#warnings.filterwarnings>`_,
+builtins from tools `like pytest <https://docs.pytest.org/en/stable/how-to/capture-warnings.html#controlling-warnings>`_ ,
```
borchero, line 19:
```suggestion
builtins from tools like `pytest <https://docs.pytest.org/en/stable/how-to/capture-warnings.html#controlling-warnings>`_ ,
```

borchero (approved):
Thanks! :)

PR #41, opened 2025-05-23
refactor: Emit FutureWarning on nullable primary key

### LINKED ISSUE #13: nullable=True is ignored if primary_key is True

AndreasAlbertQC:
# Motivation

Closes #13 

# Changes

* Added a `FutureWarning` when a user sets `nullable` and `primary_key` at the same time. Since `nullable` is otherwise quietly ignored, this should raise an error in a future release.
* refactored warning tests to make it easier to test multiple warnings

### DIFF
--- a/dataframely/columns/_base.py
+++ b/dataframely/columns/_base.py
@@ -11,7 +11,10 @@ from typing import Any
 import polars as pl
 
 from dataframely._compat import pa, sa, sa_TypeEngine
-from dataframely._deprecation import warn_nullable_default_change
+from dataframely._deprecation import (
+    warn_no_nullable_primary_keys,
+    warn_nullable_default_change,
+)
 from dataframely._polars import PolarsDataType
 from dataframely.random import Generator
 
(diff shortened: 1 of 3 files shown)


delsner (approved):
Thanks, makes sense!

dataframely/columns/_base.py
```diff
@@ -67,6 +70,10 @@ def __init__(
                 internally sets the alias to the column's name in the parent schema.
             metadata: A dictionary of metadata to attach to the column.
         """
+
+        if nullable and primary_key:
+            warn_no_nullable_primary_keys()
```
borchero, line 75: Shouldn't we already fail hard in this case? I thought we ignore the nullability property if primary key is set.

PR #44, opened 2025-05-27
feat: Add support for `time_unit` to `Datetime` column

### LINKED ISSUE #40: Support different time units in `dy.Datetime`
Brought up by @lonestike in https://github.com/Quantco/dataframely/issues/32#issuecomment-2900113037.

It would be nice to support different time units (`'us', 'ms', 'ns`) in `dy.Datetime`.

svengiegerich:
# Motivation

Closes #40 

# Changes

allows setting a time unit for datetimes

### DIFF
--- a/tests/columns/test_pyarrow.py
+++ b/tests/columns/test_pyarrow.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import pytest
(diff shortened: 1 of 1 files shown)


delsner (commented):
Thank you! Could you please add a test to check if the conversions to pyarrow work fine for all time units?

PR #49, opened 2025-06-11
refactor: Leverage Parquet metadata to serialize failure info

borchero:
# Motivation

Due to the lack of support for adding metadata to Parquet files written by polars prior to polars 1.30, we're currently writing required metadata in a rather hacky way when serializing `FailureInfo`. With polars 1.30 introducing support for parquet metadata (https://github.com/pola-rs/polars/pull/21806), we can streamline our logic.

# Changes

- Use Parquet metadata to serialize rule columns of failure info
- Bump polars version requirement to `>=1.30` (does this warrant a minor release?)

### DIFF
--- a/dataframely/failure.py
+++ b/dataframely/failure.py
@@ -102,15 +102,17 @@ class FailureInfo(Generic[S]):
         # We can read the rule columns either from the metadata of the Parquet file
         # or, to remain backwards-compatible, from the last column of the lazy frame if
         # the parquet file is missing metadata.
+        rule_columns: list[str]
+        schema_name: str
         if (meta := pl.read_parquet_metadata(source).get("dataframely")) is not None:
             metadata = json.loads(meta)
-            rule_columns: list[str] = metadata["rule_columns"]
-            schema_name: str = metadata["schema"]
-        else:  # pragma: no cover
+            rule_columns = metadata["rule_columns"]
+            schema_name = metadata["schema"]
+        else:
             last_column = lf.collect_schema().names()[-1]
             metadata = json.loads(last_column)
             rule_columns = metadata["rule_columns"]
-            schema = metadata["schema"]
+            schema_name = metadata["schema"]
             lf = lf.drop(last_column)
 
         *schema_module_parts, schema_name = schema_name.split(".")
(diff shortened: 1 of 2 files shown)


dataframely/failure.py
```diff
@@ -102,15 +98,25 @@ def scan_parquet(cls, source: str | Path | IO[bytes]) -> Self:
...
+            rule_columns: list[str] = metadata["rule_columns"]
+            schema_name: str = metadata["schema"]
+        else:  # pragma: no cover
+            last_column = lf.collect_schema().names()[-1]
+            metadata = json.loads(last_column)
+            rule_columns = metadata["rule_columns"]
+            schema = metadata["schema"]
+            lf = lf.drop(last_column)
+
+        *schema_module_parts, schema_name = schema_name.split(".")
```
delsner, line 116:
```suggestion
            schema_name = metadata["schema"]
            lf = lf.drop(last_column)

        *schema_module_parts, schema_name = schema_name.split(".")
```

Shouldn't it be like this? Have you tested if the compatibility code works?

dataframely/failure.py
```diff
@@ -102,15 +98,25 @@ def scan_parquet(cls, source: str | Path | IO[bytes]) -> Self:
...
+        if (meta := pl.read_parquet_metadata(source).get("dataframely")) is not None:
+            metadata = json.loads(meta)
+            rule_columns: list[str] = metadata["rule_columns"]
+            schema_name: str = metadata["schema"]
+        else:  # pragma: no cover
+            last_column = lf.collect_schema().names()[-1]
+            metadata = json.loads(last_column)
+            rule_columns = metadata["rule_columns"]
+            schema = metadata["schema"]
+            lf = lf.drop(last_column)
```
delsner, line 114:
```suggestion
        schema_name: str
        rule_columns: list[str]
        if (meta := pl.read_parquet_metadata(source).get("dataframely")) is not None:
            metadata = json.loads(meta)
            rule_columns = metadata["rule_columns"]
            schema_name = metadata["schema"]
        else:  # pragma: no cover
            last_column = lf.collect_schema().names()[-1]
            metadata = json.loads(last_column)
            rule_columns = metadata["rule_columns"]
            schema = metadata["schema"]
            lf = lf.drop(last_column)
```
nit: since the type hints are valid for both branches, could we maybe define them before the if-else?

delsner (commented):
Nice, great work on contributing this upstream! 🎉

PR #50, opened 2025-06-12
refactor: Remove generic parameter from collection

borchero:
# Motivation

`dy.Collection` currently allows specifying a generic parameter. The only purpose of this parameter is to provide a more concrete mapping type for the `sample` method in order to improve type annotations.

For one, this feels odd as the generic parameter influences a minor feature of the collection. It also clashes with the general programmer's notion that a collection's generic parameter influences the type of objects found in the collection.

In addition, however, this generic parameter also poses a serious problem when inheriting collections and wanting to sample both the parent and the child collection (as @MoritzPotthoffQC ran into).

For both of these reasons, I propose to get rid of the generic parameter.

# Changes

- Remove the generic parameter from `dy.Collection`, falling back to users providing simple mappings for sampling
- Copy the user input for the overrides to prevent in-place modifications in `_preprocess_sample`. This is largely orthogonal to the core change in this PR but required to turn the `Mapping` input into a `MutableMapping` (or the more concrete type `dict`)

---

Main question: Do we think this change warrants a major release? I'd expect this feature to mostly be unused.

### DIFF
--- a/dataframely/collection.py
+++ b/dataframely/collection.py
@@ -1,12 +1,11 @@
 # Copyright (c) QuantCo 2025-2025
 # SPDX-License-Identifier: BSD-3-Clause
 
-import sys
 import warnings
 from abc import ABC
 from collections.abc import Mapping, Sequence
 from pathlib import Path
-from typing import Any, Generic, Self, TypeVar, cast
+from typing import Any, Self, cast
 
 import polars as pl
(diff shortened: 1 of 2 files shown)


delsner (approved):
I also expect this feature to be mostly unused. Also, as it's just a type hint we're potentially breaking here, I don't think a major release is necessary.

PR #52, opened 2025-06-12
feat: Add `name` string property to `dy.Column`

### LINKED ISSUE #51: Accessing a columns name via `.alias` makes the type checker complain
When accessing the name of a column via `.alias`, the type checker complains, since the type is `str | None`. See the following example:
```
class MetricSchema(dy.Schema):
    name = dy.String()
    value = dy.Float64()

df.join(other_df, on=MetricSchema.name.alias)
# mypy will complain with something like
# error: Argument 2 to "join" of "Expr" has incompatible type "str | None"; expected "str"
```

Right now you either have to silence mypy with eg `# type: ignore[arg-type]` or define an extra string constant for column names. It would be nice to have a method similar to `.col()` that raises a ValueError if the column name is not set and otherwise returns the string name.

delsner:
# Motivation

Fixes #51.

# Changes

Add dynamic `name` property to `dy.Column`.

### DIFF
--- a/dataframely/columns/_base.py
+++ b/dataframely/columns/_base.py
@@ -85,7 +85,7 @@ class Column(ABC):
         self.alias = alias
         self.metadata = metadata
         # The name may be overridden by the schema on column access.
-        self.name = alias or ""
+        self._name = ""
 
     # ------------------------------------- DTYPE ------------------------------------ #
 
@@ -225,6 +225,11 @@ class Column(ABC):
         """The :mod:`pyarrow` dtype equivalent of this column data type."""
 
     # ------------------------------------ HELPER ------------------------------------ #
+
+    @property
+    def name(self) -> str:
+        """Get the name of the column in a schema."""
+        return self._name
 
     @property
     def col(self) -> pl.Expr:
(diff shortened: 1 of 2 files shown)


borchero (commented):
I had one thought: could we somehow make `.col` and `.alias` (or something like `.name`) be populated dynamically? Specifically, we could overwrite `__getattr__` (or `__getattribute__`, I always forget what's the right one) of `Schema` and set the column name there, i.e.

```python
def __getattr__(self, name: str) -> Any:
    val = super().__getattr__(name)
    if isinstance(val, Column):
        val.name = val.alias or name
    return val
```

That would also allow fixing the problem that we had discussed previously where you can't set the column of a schema based on the column of another schema, i.e. something like

```python
column = OtherSchema.column_with_other_name
```

dataframely/columns/_base.py
```diff
@@ -85,6 +84,8 @@ def __init__(
         self.check = check
         self.alias = alias
         self.metadata = metadata
+        # The name may be overridden by the schema on column access.
+        self.name = alias or ""
```
borchero, line 88: I think it would make sense for `name` to be only truly available if the column is accessed via the schema. Hence, I'd probably set this to `""` to begin with and disregard the alias. The fact that the alias is available here does not make sense conceptually IMO, it's simply the only place where one can specify the name of a column with a non-Python-identifier-name
```suggestion
        self.name = ""
```

borchero: Also, would it make sense to make this `_name` and expose another property `name`? This would allow documenting it and making it obvious that it is similar to `col`

delsner: c8570b5

PR #57, opened 2025-06-15
feat: Introduce (de-)serialization for schemas

borchero:
# Motivation

When serializing data frames to parquet, schema information beyond the simple polars schema is currently lost. However, persisting the schema alongside has multiple potential benefits:

- When reading the parquet file, schema information can be used to check whether the data adheres to the current schema without having to validate.
- Third-party tools can use advanced schema information to optimize operations

Since polars 1.30, the parquet writer supports writing file-level metadata. Naturally, dataframely schemas could be serialized as such.

This PR is the first step towards doing that: it introduces (de-)serialization for schemas from & to JSON.

# Changes

- Introduce `Schema.serialize` and `dy.deserialize_schema` to serialize and de-serialize schemas to and from JSON

### DIFF
--- a/dataframely/__init__.py
+++ b/dataframely/__init__.py
@@ -51,7 +51,7 @@ from .functional import (
     filter_relationship_one_to_at_least_one,
     filter_relationship_one_to_one,
 )
-from .schema import Schema
+from .schema import Schema, load_schema
 
 __all__ = [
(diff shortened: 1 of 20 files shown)


AndreasAlbertQC (approved):
Nice!

Issue #59, opened 2025-06-16
Eager evaluation of methods by `dy.rule` can lead to `NameErrors`


FrithjofWinkelmannQC:
Because `@dy.rule` calls the annotated method immediately and thus the containing module might not be initialized, `NameErrors` can occur if not yet initialized values are referenced. It would also be great to be able to annotate `@dy.rule`s with `@classmethod`, but this is also not possible right now. see the example below:
```
class MyComplexSchema(dy.Schema):
    a = dy.Int64()
    b = dy.Int64()

    @dy.rule()
    def b_greater_constant() -> pl.Expr:
        return MyComplexSchema.b.col > SOME_CONSTANT_DEFINED_LATER # NameError: name 'SOME_CONSTANT_DEFINED_LATER' is not defined

    @dy.rule()
    def b_greater_a() -> pl.Expr:
        return MyComplexSchema.b.col > MyComplexSchema.a.col # NameError: name 'MyComplexSchema' is not defined

    # This causes  `TypeError: 'classmethod' object is not callable`
    @dy.rule() # error: Argument 1 has incompatible type "Callable[[Any], Expr]"; expected "Callable[[], Expr]"
    @classmethod
    def b_greater_a_cls_method(cls) -> pl.Expr:
        return cls.b.col > cls.a.col

SOME_CONSTANT_DEFINED_LATER = 5
```

PR #60, opened 2025-06-16
feat: Add lazy rules

FrithjofWinkelmannQC:
# Motivation

This is a rough draft to solve #59.

# Changes

This adds another annotation `@dy.lazy_rule()` that works like `@dy.rule()` but defers evaluation of the wrapped method until a dataframe is actually verified.

As @borchero pointed out, we cant defer the eager evaluation of `@dy.rule` without dropping the check that rules return expressions of dtype boolean. `@dy.lazy_rule()` was introduced as an extra annotation in order to keep error messages on `@dy.rule` the same.

### DIFF
--- a/dataframely/_rule.py
+++ b/dataframely/_rule.py
@@ -80,6 +80,26 @@ def rule(
 def lazy_rule(
     *, group_by: list[str] | None = None
 ) -> Callable[[LazyValidationFunction], Rule]:
+    """Mark a function as a rule to evaluate during validation.
+
+    This does the same as `rule`, but allows for lazy evaluation of rules. This means the
+    dtype of the returned expression is not validated to be a boolean expression! Use this
+    method if you need to access the schema class in your rule function, want to use
+    `@classmethod` on the rule or need to access constants defined after the schema.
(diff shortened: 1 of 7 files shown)


dataframely/_rule.py
```diff
@@ -67,14 +95,30 @@ def decorator(validation_fn: ValidationFunction) -> Rule:
 # ------------------------------------------------------------------------------------ #
 
 
-def with_evaluation_rules(lf: pl.LazyFrame, rules: dict[str, Rule]) -> pl.LazyFrame:
+def _call_lazy_rule(
```
FrithjofWinkelmannQC, line 98: Somewhere around this method we should potentially check if the returned expression actually has dtype boolean, as we skip this check for lazy rules on schema initialization.

borchero:
Thanks for being proactive here @FrithjofWinkelmannQC, this was very useful inspiration! We decided to keep the API as simple as it is and just remove the currently printed error message when rules are implemented incorrectly (see #64). We will add a better error message in a follow-up PR via another mechanism.

PR #62, opened 2025-06-17
feat: Allow semantic comparison of collections

AndreasAlbertQC:
# Motivation

Same as https://github.com/Quantco/dataframely/pull/55 but for `Collection`s instead of `Schema`s. Preparation for serialization of `Collection` definitions.

# Changes

* Implemented `Collection.matches`, which semantically compares the present collection with another one and returns a boolean indicating whether the two Collections are equivalent,.

### DIFF
--- a/dataframely/_base_schema.py
+++ b/dataframely/_base_schema.py
@@ -149,9 +149,9 @@ class SchemaMeta(ABCMeta):
     def __repr__(cls) -> str:
         parts = [f'[Schema "{cls.__name__}"]']
         parts.append(textwrap.indent("Columns:", prefix=" " * 2))
-        for name, col in cls.columns().items():  # type: ignore
+        for name, col in cls.columns().items():
             parts.append(textwrap.indent(f'- "{name}": {col!r}', prefix=" " * 4))
-        if validation_rules := cls._schema_validation_rules():  # type: ignore
+        if validation_rules := cls._schema_validation_rules():
             parts.append(textwrap.indent("Rules:", prefix=" " * 2))
             for name, rule in validation_rules.items():
                 parts.append(textwrap.indent(f'- "{name}": {rule!r}', prefix=" " * 4))


dataframely/_rule.py
```diff
@@ -91,6 +91,13 @@ def rule(*, group_by: list[str] | None = None) -> Callable[[ValidationFunction],
         rules. By default, any rule that evaluates to ``null`` because one of the
         columns used in the rule is ``null`` is interpreted as ``true``, i.e. the row
         is assumed to be valid.
+
```
borchero, line 94:
```suggestion
```

dataframely/collection.py
```diff
@@ -225,6 +226,58 @@ def sample(
...
+
+        # Computational graph of filter logic must match
+        # Evaluate on empty dataframes
+        empty_left = cls.create_empty()
+        empty_right = other.create_empty()
+
+        def filter_logic_matches(filter_left: Filter, filter_right: Filter) -> bool:
+            lf_left = filter_left.logic(empty_left)
+            lf_right = filter_right.logic(empty_right)
+            return lf_left.explain() == lf_right.explain()
```
borchero, line 271: Can we use https://docs.pola.rs/api/python/stable/reference/lazyframe/api/polars.LazyFrame.serialize.html? `explain` runs optimization so it is considerably more expensive. `serialize` also seems to be more structured.

tests/collection/test_matches.py
```diff
@@ -0,0 +1,95 @@
...
+    # Two schemas that do not match
+    class MyIntSchema(dy.Schema):
+        foo = dy.Integer()
+
+    class MyStringSchema(dy.Schema):
+        foo = dy.String()
+
+    assert not MyIntSchema.matches(MyStringSchema), (
+        "Test schemas must not match for test setup to make sense"
+    )
```
borchero, line 46: I don't think you need this, this functionality should really be covered by the schema tests.

borchero (commented):
Thanks! Just a few comments on implementational details :)

tests/collection/test_matches.py
AndreasAlbertQC: I did not mean this as a test that `Schema.matches` works (which indeed, should happen in the schema tests), I meant it as a sanity check for the setup of this specific test. The logic of the test is meant to be "if the schemas are different, collection-level comparison should fail", so I want to verify that I indeed created a setup that fulfills the first half of that sentence.

dataframely/collection.py
```diff
@@ -225,6 +226,65 @@ def sample(
...
+
+            # Computational graph of filter logic must match
+            # Evaluate on empty dataframes
+            empty_left = cls.create_empty()
+            empty_right = other.create_empty()
+
+            for name in filters_lhs:
+                lhs = filters_lhs[name].logic(empty_left)
+                rhs = filters_rhs[name].logic(empty_right)
+                if not lhs.serialize(format="json") == rhs.serialize(format="json"):
```
borchero, line 282:
```suggestion
                if lhs.serialize(format="json") != rhs.serialize(format="json"):
```

dataframely/collection.py
```diff
@@ -225,6 +226,65 @@ def sample(
...
+            # Member attributes must match
+            for name in members_lhs:
+                lhs = asdict(members_lhs[name])
+                rhs = asdict(members_rhs[name])
+                for attr in lhs.keys() | rhs.keys():
+                    if attr == "schema":
+                        if not lhs[attr].matches(rhs[attr]):
+                            return False
+                    else:
+                        if not lhs[attr] == rhs[attr]:
```
borchero, line 262:
```suggestion
                        if lhs[attr] != rhs[attr]:
```

borchero (approved):
Looks great, thank you! Just two comments to apply De Morgan :smile:

PR #63, opened 2025-06-17
feat: Implement dunder repr for collection, schema, column and rule

MoritzPotthoffQC:
# Motivation

We would like to make schemas printable. 

# Changes

Implemented `__repr__` for `Collection`, `Schema`, `Column` and `Rule`

### DIFF
--- a/dataframely/_base_collection.py
+++ b/dataframely/_base_collection.py
@@ -261,14 +261,14 @@ class CollectionMeta(ABCMeta):
             )
         if filters := cls._filters():  # type: ignore
             parts.append(textwrap.indent("Filters:", prefix=" " * 2))
-        for name, member in filters.items():
-            parts.append(textwrap.indent(f'- "{name}":', prefix=" " * 4))
-            parts.append(
-                textwrap.indent(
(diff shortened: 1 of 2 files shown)


tests/collection/test_repr.py
```diff
@@ -0,0 +1,37 @@
...
+            - "member_a": MySchema(optional=False, ignored_in_filters=False, inline_for_sampling=False)
+            - "member_b": MySchema(optional=False, ignored_in_filters=False, inline_for_sampling=False)
+          Filters:
+            - "member_a_member_b_one_to_one":
+              INNER JOIN:
+              LEFT PLAN ON: [col("a")]
+                DF ["a"]; PROJECT */1 COLUMNS
+              RIGHT PLAN ON: [col("a")]
+                DF ["a"]; PROJECT */1 COLUMNS
+              END INNER JOIN""")
```
MoritzPotthoffQC, line 37: I also added filters not. Lacking a better idea, I print the explanation of the plan for the resulting lazy frame. I also tried `serialize` from the comment on #62, but that is not very readable even in json format.

borchero (approved):
Looks great, thank you!

PR #64, opened 2025-06-17
refactor: Refactor (group-)rule to be lazily evaluated

delsner:
# Motivation

Addresses #59.

# Changes

We don't want to introduce a separate decorator as done in #60 with `dy.lazy_rule`. 
Instead this PR instead lazily evaluates (group-)rules and gets rid of the column check which requires all symbols used inside the rule expression to be defined (including the schema class itself).


borchero (approved):
Thanks! 🚀

PR #67, opened 2025-06-25
fix: Add `drop_non_unique` parameter to `filter_relationship_one_to_one`

delsner:
# Motivation

`filter_relationship_one_to_one` does only check for a 1:1 relationship in case both data frames are unique w.r.t. the provided join key. Also, the docstrings are incorrect as the join columns cannot be inferred.

# Changes

* Add `drop_non_unique` to allow filtering for 1:1 relationships in case join columns do not uniquely identify rows
* Fix docstrings

### DIFF
--- a/dataframely/functional.py
+++ b/dataframely/functional.py
@@ -23,22 +23,43 @@ C = TypeVar("C", bound=BaseCollection)
 
 # --------------------------------- RELATIONSHIP 1:1 --------------------------------- #
 
+LEN_COLUMN = "__len__"
+
 
 def filter_relationship_one_to_one(
     lhs: LazyFrame[S] | pl.LazyFrame,
     rhs: LazyFrame[T] | pl.LazyFrame,
     /,
     on: str | list[str],
+    *,
+    drop_non_unique: bool = False,
 ) -> pl.LazyFrame:
     """Express a 1:1 mapping between data frames for a collection filter.
 
     Args:
         lhs: The first data frame in the 1:1 mapping.
         rhs: The second data frame in the 1:1 mapping.
-        on: The columns to join the data frames on. If not provided, the join columns
-            are inferred from the mutual primary keys of the provided data frames.
+        on: The columns to join the data frames on.
+        drop_non_unique: If `True`, drop non-unique rows from both data frames.
+            This is useful when the join columns do not already uniquely identify rows.
     """
-    return lhs.join(rhs, on=on)
+    if drop_non_unique:
+        return (
+            lhs.group_by(on)
+            .len(LEN_COLUMN)
+            .filter(pl.col(LEN_COLUMN) == 1)
+            .drop(LEN_COLUMN)
+            .join(
+                rhs.group_by(on)
+                .len(LEN_COLUMN)
+                .filter(pl.col(LEN_COLUMN) == 1)
+                .drop(LEN_COLUMN),
+                on=on,
+                how="inner",
+            )
+        )
+    else:
+        return lhs.join(rhs, on=on)
 
 
 # ------------------------------- RELATIONSHIP 1:{1,N} ------------------------------- #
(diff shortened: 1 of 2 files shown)


delsner:
@borchero wdyt of this fix? One could argue that these functions are only to be used inside `dy.filter` where you would typically pass the common primary key as join key, but I still find it unintuitive that we don't actually filter for a 1:1 relationship.

dataframely/functional.py
```diff
@@ -23,22 +23,43 @@
...
 ) -> pl.LazyFrame:
     """Express a 1:1 mapping between data frames for a collection filter.
 
     Args:
         lhs: The first data frame in the 1:1 mapping.
         rhs: The second data frame in the 1:1 mapping.
-        on: The columns to join the data frames on. If not provided, the join columns
-            are inferred from the mutual primary keys of the provided data frames.
+        on: The columns to join the data frames on.
+        drop_non_unique: If `True`, drop non-unique rows from both data frames.
```
robinholzi, line 43: Nice, just one minor thought on naming.
Filter rules in dataframely are used in `.validate` and `.filter` calls. In the former they drop rows that `should lead to failure` in the latter they will `drop rows` from the "good" samples (putting them into the failures dataframe). I would only refer to `dropping` in the latter case. Hence, I would not call the argument `drop` inside the `filter` rule (helper function). Maybe `allow_non_unique`?

robinholzi (approved):

AndreasAlbertQC:
@delsner I think your change is good, but I wonder if we need to do something else to mitigate confusing results if the dataframes are not unique on the join keys already. If I don't set the new flag, I will get nonsensical results, right?
Possible solutions:

1. check for uniqueness and if its violated, do one of the following:
  1.1 hard error
  1.2 warning
  1.3 automatically activate the new flag plus warning
2. Don't check anything, and document that you will get nonsensical results if you don't ensure that the keys are unique already, so the user has to take care of it.

I am hesitating between 1.3 and 2.

delsner:
> If I don't set the new flag, I will get nonsensical results, right?

Potentially, yes. 

> I am hesitating between 1.3 and 2.

2 is easy, I can just update the docstring.
IMO 1.3 is the right way to do it, but this is a breaking change. I think hardly anyone is using this feature since it's undocumented, so I'd also be fine to go with 1.3. Wdyt?

Any thoughts @borchero?

AndreasAlbertQC:
@delsner I'd be happy with counting option 1.3 as a bug fix bc I think the current behavior generates incorrect output data

borchero:
I made the current choice deliberately as additional uniqueness validations are _really_ expensive. Hence, my initial thought would be to go for (2).

Nevertheless, it potentially makes sense to make this behavior opt-in, i.e. require the user to set some kind of flag to skip the additional validation step.

However, I'm not sure I love the extension of the `filter_relationship_one_to_one` method; we're somehow evaluating a "many-to-many" relationship here, don't we? 🤔

delsner:
We decided to rename these functions to `require_*` and also will update the `one_to_at_least_one` function. We'll also rename the parameter to `is_unique` and change its default to `False`.

tests/functional/test_relationships.py
```diff
@@ -29,7 +29,7 @@ class EmployeeSchema(dy.Schema):
 
 @pytest.fixture()
 def departments() -> dy.LazyFrame[DepartmentSchema]:
-    return DepartmentSchema.cast(pl.LazyFrame({"department_id": [1, 2]}))
```
AndreasAlbertQC: AFAICT `departments` is always the `left` dataframe when we call `require_relationship_one_to_one` in the tests below. Don't we also need a test where the `left` dataframe is not unique in the join column?

dataframely/functional.py
```diff
@@ -23,41 +23,86 @@
...
+    Returns:
+        A data frame representing the inner join of the two inputs on the specified
+        columns, filtered to ensure a 1:1 relationship.
     """
+    if filter_unique:
+        return (
+            lhs.group_by(on)
+            .len(LEN_COLUMN)
+            .filter(pl.col(LEN_COLUMN) == 1)
+            .drop(LEN_COLUMN)
```
borchero: Why are we not using `unique` here? 👀 
```suggestion
            lhs.unique(on)
```

borchero: Ah yes, absolutely, we need the `keep="none"`!

PR #77, opened 2025-07-06
feat: Allow to efficiently read and write collections

borchero:
# Motivation

This should be the last serialization-related PR for now and completes the full support for proper persistence operations :smile:

**Caveat:** This method _might_ be considered breaking as `Collection.scan_parquet` previously never ran any validation. Even though this was documented, I would rather consider this a bug...


dataframely/collection.py
```diff
@@ -620,84 +625,230 @@ def serialize(cls) -> str:
...
+        with open(path / "schema.json", "w") as f:
+            f.write(self.serialize())
+
+        member_schemas = self.member_schemas()
+        for key, lf in self.to_dict().items():
+            member_schemas[key].sink_parquet(
+                lf,  # type: ignore
+                path / f"{key}.parquet",
+                **kwargs,
+            )
```
AndreasAlbertQC, line 696: did you consider dispatching to an internal `_to_parquet(directory: Path, sink: bool, **kwargs)` here to unify the sink and write methods, since they are mostly identical code?

dataframely/collection.py
```diff
@@ -620,84 +625,230 @@ def serialize(cls) -> str:
...
+        cls,
+        directory: Path,
+        validate: bool | Literal["auto"] = "auto",
+    ) -> bool:
+        # First, we check whether the path provides the serialization of the collection.
+        # If it does, we check whether it matches this collection. If it does, we assume
+        # that the data adheres to the collection and we do not need to run validation.
+        if (json_serialization := directory / "schema.json").exists():
+            with json_serialization.open("r") as f:
+                metadata = f.read()
```
AndreasAlbertQC, line 829:
```suggestion            
            metadata = json_serialization.read_text()
```
no biggie, but why not?

AndreasAlbertQC (approved):

dataframely/collection.py
```diff
@@ -620,84 +625,230 @@ def serialize(cls) -> str:
...
         members which are not provided in the current collection.
 
+        In addition, one JSON file named ``schema.json`` is written, serializing the
+        collection's definition for fast reads.
+
         Args:
             directory: The directory where the Parquet files should be written to. If
                 the directory does not exist, it is created automatically, including all
                 of its parents.
+            kwargs: Additional keyword arguments passed directly to
```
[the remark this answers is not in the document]
borchero: Decided to properly support partitioning.

PR #84, opened 2025-07-09
refactor: Change `Literal[None]` to `None`

GameRoMan:


GameRoMan:
> Hey @GameRoMan, thanks for your PR! Can you provide more information whether this is semantically equivalent? I never thought about using `: None` as a type annotation for a parameter before.

Hi, types `None` and `Literal[None]` are [exactly equivalent](https://typing.python.org/en/latest/spec/literal.html#:~:text=Note%3A%20Since%20the%20type%20None%20is%20inhabited%20by%20just%20a%20single%20value%2C%20the%20types%20None%20and%20Literal%5BNone%5D%20are%20exactly%20equivalent.%20Type%20checkers%20may%20simplify%20Literal%5BNone%5D%20into%20just%20None.)

borchero (approved):
Thanks for the link, looks good! 🚀

PR #86, opened 2025-07-10
refactor: Move `Validation` to public API

delsner:
# Motivation

If one wants to override functions from the public API of `dy.Schema` or `dy.Collection`, the `Validation` literal type should also be exposed.

# Changes

Add `Validation` to public API.


borchero:
> If one wants to override functions from the public API 

Why do you want to do that though? :smile:

delsner:
xref https://github.com/Quantco/dataframely/issues/87

PR #88, opened 2025-07-11
build: Support Python 3.10

delsner:
# Motivation

Enable projects which cannot upgrade from 3.10 yet to use dataframely.

# Changes

Support Python 3.10.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -86,6 +86,7 @@ pre-commit-run = "pre-commit run -a"
 
 [feature.py310.dependencies]
 python = "3.10.*"
+typing-extensions = "*"
 
 [feature.py311.dependencies]
 python = "3.11.*"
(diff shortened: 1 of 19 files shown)


pixi.toml
```diff
@@ -84,6 +84,9 @@ typos = "*"
 pre-commit-install = "pre-commit install"
 pre-commit-run = "pre-commit run -a"
 
+[feature.py310.dependencies]
+python = "3.10.*"
```
borchero, line 88: Don't we need to add `typing-extensions` as additional dependency for `py310`?

pyproject.toml
borchero, line 1: We need to add `typing-extensions` as dependency conditional on `python_version<=3.10`

borchero (commented):
Let's do it, just two comments regarding dependency management

pixi.toml
delsner: [fc59ee6](https://github.com/Quantco/dataframely/pull/88/commits/fc59ee69eb2dd7c9d84e8e5c303df874771f01d9)

gmhewett:
:rocket:

PR #89, opened 2025-07-11
fix: Only warn about nullability on non-primary-key columns

AndreasAlbertQC:
# Motivation

The current FutureWarning also triggers on primary key columns, which cannot be nullable anyway.

# Changes

* Guarded the warning such that it is not emitted on primary key columns


borchero (approved):

AndreasAlbertQC:
@borchero there was still a fix necessary after you reviewed

PR #91, opened 2025-07-15
fix: Conditionally import typing-extensions (#90)

### LINKED ISSUE #90: ImportError caused by typing-extensions
Hi,

I've noticed that since support for Python 3.10 was added (727cc55c11ca1f5d813139f6a153fdaac63c62be), typing-extensions has been listed in the pyproject.toml as a dependency when using Python <= 3.10. 

This is causing an ImportError crash for me on Python 3.13 because typing-extensions doesn't get installed alongside dataframely, but it is imported unconditionally (e.g. [see here](https://github.com/Quantco/dataframely/blob/7a12bc6a310cc7641d9cb875d00b05f822c3b9f7/dataframely/_base_collection.py#L14))

To avoid this, typing-extensions can be imported conditionally, eg:

```python
import sys

if sys.version_info >= (3, 10):
    from typing import Self
else:
    from typing_extensions import Self
```

I'm happy to make a PR for this if you like.

Thanks

jhall-bp:
# Motivation

This PR is to resolve #90. typing-extensions is listed as a dependency when using Python 3.10 or earlier, but is always imported from leading to ImportErrors when using dataframely on newer Python versions.

# Changes

I have changed imports from the typing_extensions module that were added to support Python 3.10 from:

```python
from typing_extensions import X
```

to:

```python
import sys
if sys.version_info > (3, 10):
    from typing import X
else:
    from typing_extensions import X
```

This means that Unpack, NotRequired, and Self are imported from typing when using 3.11 or newer and from typing_extensions otherwise.

### DIFF
--- a/dataframely/_base_collection.py
+++ b/dataframely/_base_collection.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import sys
 import textwrap
 import typing
 from abc import ABCMeta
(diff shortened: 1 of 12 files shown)


borchero (approved):
Thanks @jhall-bp! Turns out `> (3, 10)` is different to `>= (3, 11)` since `(3, 10, 8) > (3, 10)` evaluates to `True`, so, I changed this myself to make a release asap :smile:

PR #107, opened 2025-08-01
refactor: Package collection metadata into parquets

AndreasAlbertQC:
# Motivation

When serializing collections, we currently rely on creating a `schema.json` file to store the serialized metadata. This approach has the downside that we need to track this extra object, which is structurally different from the parquet files we write. In individual dataframes, we have already unified storage of data and metadata by packaging the metadata into the parquet. This PR uses the same approach to package the metadata of the collection into the members parquet files.

This change is currently not backward compatible. If we decide to go this way, I'd suggest that we add the functionality to fall back to reading `schema.json` files if they exist. I think this should be reasonably straightforward and would establish backward compatibility for reads of previously written files.

# Changes

* Removed the notion `schema.json`.
* Refactored the parquet serialization code to package the metadata information about collections into the parquet metadata.
* When reading back the collection, added checks to ensure that all parquets have the same collection meta data. If they do not, this is treated as if no metadata was found.

### DIFF
--- a/dataframely/collection.py
+++ b/dataframely/collection.py
@@ -987,7 +987,7 @@ def read_parquet_metadata_collection(
         source: Path to a parquet file or a file-like object that contains the metadata.
 
     Returns:
-        The schema that was serialized to the metadata or ``None`` if no schema metadata
+        The schema that was serialized to the metadata or ``None`` if no collection metadata
         is found.
     """
     metadata = pl.read_parquet_metadata(source)
--- a/dataframely/collection.py
+++ b/dataframely/collection.py
@@ -841,8 +841,11 @@ class Collection(BaseCollection, ABC):
                     if scan
                     else pl.read_parquet(source_path, **kwargs).lazy()
                 )
-                parquet_collection_type = read_parquet_metadata_collection(source_path)
-                collection_types.add(parquet_collection_type)
+                if source_path.is_file():
+                    collection_types.add(read_parquet_metadata_collection(source_path))
+                else:
+                    for file in source_path.glob("**/*.parquet"):
+                        collection_types.add(read_parquet_metadata_collection(file))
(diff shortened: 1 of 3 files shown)


dataframely/collection.py
```diff
@@ -968,3 +989,39 @@ def _extract_keys_if_exist(
...
+def read_parquet_metadata_collection(
+    source: str | Path | IO[bytes] | bytes,
+) -> type[Collection] | None:
+    """Read a dataframely Collection type from the metadata of a parquet file.
+
+    Args:
+        source: Path to a parquet file or a file-like object that contains the metadata.
+
+    Returns:
+        The schema that was serialized to the metadata or ``None`` if no collection metadata
```
AndreasAlbertQC, line 1003:
```suggestion
        The collection that was serialized to the metadata or ``None`` if no collection metadata
```

tests/collection/test_read_write_parquet.py
```diff
@@ -29,7 +30,16 @@ def _write_parquet(collection: dy.Collection, path: Path, lazy: bool) -> None:
         collection.sink_parquet(path)
     else:
         collection.write_parquet(path)
-    (path / "schema.json").unlink()
+
+    def _delete_meta(file: Path) -> None:
```
AndreasAlbertQC, line 34:
```suggestion
    def _delete_meta(file: Path) -> None:
        """Overwrite a parquet file with the same data, but without metadata"""
```

dataframely/collection.py
```diff
@@ -903,6 +925,24 @@ def _validate_input_keys(cls, data: Mapping[str, FrameType], /) -> None:
             )
 
 
+def read_parquet_metadata_collection(
```
borchero, line 928: Are you moving the `read_parquet_metadata` method we discussed to another PR? :smile:

PR #109, opened 2025-08-06
refactor: Move IO specifics out of `Schema`, `Collection`

AndreasAlbertQC:
# Motivation

The current implementation of parquet-based storage is strongly coupled to the implementation of the `Schema` and `Collection` classes. This makes it hard to reuse parts of the logic that would also be useful for other future storage backends. To prepare the implementation of such future storage backends, this PR refactors the current logic to introduce a clearer interface for storage backends.

# Changes

* Introduced the notion of a `StorageBackend`, i.e. a piece of code encapsulating a way of storing data and metadata.
* Refactored the parquet storage logic to use the `StorageBackend` interface


dataframely/schema.py
```diff
@@ -893,6 +896,56 @@ def _requires_validation_for_reading_parquet(
...
+        validation: Validation,
+        lazy: Literal[False],
+        **kwargs: Any,
+    ) -> DataFrame[Self]: ...
+
+    @classmethod
+    def _read(
+        cls, io: StorageBackend, validation: Validation, lazy: bool, **kwargs: Any
+    ) -> LazyFrame[Self] | DataFrame[Self]:
+        read_function = io.scan_frame if lazy else io.read_frame
```
borchero, line 932: This feels weird, can we simply do an `if` block below to assign `df` and `serialized_schema`?

AndreasAlbertQC:
ping @borchero

dataframely/_serialization.py
```diff
@@ -118,3 +121,367 @@ def object_hook(self, dct: dict[str, Any]) -> Any:
...
             case _:
                 raise TypeError(f"Unknown type '{dct['__type__']}' in JSON data.")
+
+
+SerializedSchema = str
+SerializedCollection = str
+
+
+class StorageBackend(ABC):
+    """Base class for IO managers.
```
borchero, line 131: The description conflicts a little with the classname. Can we align?

dataframely/_serialization.py
```diff
@@ -118,3 +121,367 @@ def object_hook(self, dct: dict[str, Any]) -> Any:
...
+        """This method stores frames as individual parquet files.
+
+        Args:
+            lf: LazyFrame containing the data to be stored.
+            kwargs: The "file" kwarg is required to specify where data is stored.
+                It should point to a parquet file. If hive partitioning is used,
+                it should point to a directory.
+                The "metadata" kwarg is supported to pass a dictionary of parquet
+                metadata.
+                Additional keyword arguments are passed to polars.
```
borchero, line 290: I'm not sure if all this documentation is required, it should be self-explanatory from the superclass. I would even do away with the summary line here and simply expect maintainers (this is not user-facing) to read the class docs (or the little code that's in here).

dataframely/schema.py
```diff
@@ -876,6 +879,68 @@ def _requires_validation_for_reading_parquet(
...
+        if cls._requires_validation_for_reading_parquet(
+            deserialized_schema, validation, source=str(source)
+        ):
+            validated = cls.validate(lf, cast=True)
+            if lazy:
+                return validated.lazy()
+            else:
+                return validated
+
+        casted = cls.cast(lf)
```
borchero, line 938: Did we always cast, no matter what? 🤔

dataframely/_serialization.py
```diff
@@ -118,3 +121,367 @@ def object_hook(self, dct: dict[str, Any]) -> Any:
...
+        for key, lf in dfs.items():
+            destination = (
+                path / key if "partition_by" in kwargs else path / f"{key}.parquet"
+            )
+            self.sink_frame(
+                lf,
+                serialized_schema=serialized_schemas[key],
+                file=destination,
+                **kwargs,
+            )
```
borchero, line 378: Potentially for a follow-up PR: `sink_parquet` has a `lazy=True` parameter (see https://docs.pola.rs/api/python/stable/reference/api/polars.LazyFrame.sink_parquet.html). This allows us to create a separate lazy frame for writing all collection members, then call `collect_all` for maximum speedup ;)

dataframely/_serialization.py
```diff
@@ -118,3 +121,367 @@ def object_hook(self, dct: dict[str, Any]) -> Any:
                 )
             case _:
                 raise TypeError(f"Unknown type '{dct['__type__']}' in JSON data.")
+
+
+SerializedSchema = str
+SerializedCollection = str
+
+
+class StorageBackend(ABC):
```
borchero, line 130: Potentially, we could think about creating a new subdirectory for this storage stuff; it's a lot of code for one file already 👀

borchero (commented):
Just some cosmetic comments, thanks! 🚀

dataframely/schema.py
AndreasAlbertQC: If validation is required, we already return inside the if-block right above this. So we only cast if validation is not required, which seems sensible to me.

borchero (approved):
Let's go :D

PR #112, opened 2025-08-07
feat: Add `Categorical` column type

### LINKED ISSUE #101: Add categorical datatype
`dataframely` doesn't appear to support `pl.categorical` datatype. Could this be added ?

borchero:
# Motivation

With polars 1.32 removing the global string cache, this makes sense to support. Fixes #101.


delsner (approved):
LGTM!

PR #114, opened 2025-08-08
fix: Be more lenient when reading metadata

borchero:
# Motivation

A bump in the polars version can lead to schema metadata being unreadable. Currently, this causes an exception. Instead, we should simply handle this just like when no schema is available.

In addition, this PR unifies the way that expressions and lazy frames are serialized and uses the binary format for both (the JSON format is not available for lazy frames, so, this is the only possible choice if one wants to use the same serialization format for both).

### DIFF
--- a/dataframely/_serialization.py
+++ b/dataframely/_serialization.py
@@ -46,7 +46,7 @@ class SchemaJSONEncoder(JSONEncoder):
             case pl.Expr():
                 return {
                     "__type__": "expression",
-                    "value": obj.meta.serialize(format="json"),
+                    "value": base64.b64encode(obj.meta.serialize()).decode("utf-8"),
                 }
             case pl.LazyFrame():
                 return {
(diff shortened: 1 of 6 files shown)


dataframely/_serialization.py
```diff
@@ -87,8 +87,13 @@ def object_hook(self, dct: dict[str, Any]) -> Any:
             case "tuple":
                 return tuple(dct["value"])
             case "expression":
-                data = BytesIO(cast(str, dct["value"]).encode("utf-8"))
-                return pl.Expr.deserialize(data, format="json")
+                value_str = cast(str, dct["value"]).encode("utf-8")
+                if value_str.startswith(b"{"):
+                    data = BytesIO(value_str)
+                    return pl.Expr.deserialize(data, format="json")
```
delsner, line 93: If this is mainly for backwards-compatibility, should we add a comment?

tests/test_serializtion.py
delsner, line 1: nit: typo in filename

delsner (approved):

PR #120, opened 2025-08-12
feat: Allow specifying custom base collection when creating a dynamic collection

MoritzPotthoffQC:
# Motivation

Currently, all dynamic collections are based directly on `dy.Collection`. It would be nice to allow basing them on a custom collection type so that the dynamic collection inherits whatever additional logic exists there.  

# Changes

- Added base collection as an argument, defaulting to the current behavior.

### DIFF
--- a/dataframely/testing/factory.py
+++ b/dataframely/testing/factory.py
@@ -36,6 +36,7 @@ def create_collection(
     schemas: dict[str, type[Schema]],
     filters: dict[str, Filter] | None = None,
     *,
+    collection_base_class: type = Collection,
     annotation_base_class: type = LazyFrame,
 ) -> type[Collection]:
     return create_collection_raw(
@@ -45,6 +46,7 @@ def create_collection(
             for name, schema in schemas.items()
         },
         filters=filters,
+        collection_base_class=collection_base_class,
     )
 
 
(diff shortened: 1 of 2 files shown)


dataframely/testing/factory.py
```diff
@@ -33,29 +33,49 @@ def create_schema(
 
 def create_collection(
     name: str,
-    schemas: dict[str, type[Schema]],
+    schemas: dict[str, type[Schema]] | None = None,
```
borchero, line 36: Why would I want to ever call this with `None`? If I already have a collection with all the schemas I want, I don't need this function, do I?

tests/test_factory.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@pytest.fixture()
+def member() -> pl.LazyFrame:
+    return pl.LazyFrame({"a": [0, 1, 2, 3, 10]})
+
+
+@pytest.fixture()
+def member2() -> pl.LazyFrame:
+    return pl.LazyFrame({"a": [0, 1, 2, 3, 10], "b": ["a", "b", "c", "d", "e"]})
```
borchero, line 45: I'm not a fan of data in fixtures as changing the fixture for a test requires updating all other tests. Can we just hardcode these in the tests themselves and duplicate them?

borchero (approved):
Thanks! Two small comments :)

dataframely/testing/factory.py
MoritzPotthoffQC: You could use this to only add new filters. Up to you though if we want to keep it

borchero: Mh yeah, you could provide an empty dict regardless 😄 I'd _rather_ make it non-optional but no strong opinion, you decide 🤪

MoritzPotthoffQC: Fair enough, made it non-optional now

PR #127, opened 2025-08-14
feat: Add `join` for Collections

MoritzPotthoffQC:
# Motivation

When working with collections, it is often helpful to filter all members by the common primary key or remove certain entries this way. 

# Changes

- Added `join` method to `Collection` that allows to semi- or anti-join a collection onto a dataframe containing entries for the common primary key
- Added tests

### DIFF
--- a/dataframely/collection.py
+++ b/dataframely/collection.py
@@ -565,23 +565,32 @@ class Collection(BaseCollection, ABC):
         self,
         primary_keys: pl.LazyFrame,
         how: Literal["semi", "anti"] = "semi",
+        maintain_order: Literal["none", "left"] = "none",
     ) -> Self:
         """Filter the collection by joining onto a data frame containing entries for the
-        common primary key columns whose respective rows should be kept or removed in the
-        collection members.
+        common primary key columns whose respective rows should be kept or removed in
+        the collection members.
 
         Args:
             primary_keys: The data frame to join on. Must contain the common primary key
                 columns of the collection.
             how: The join strategy to use. Like in polars, `semi` will keep all rows
                 that can be found in `primary_keys`, `anti` will remove them.
+            maintain_order: The `maintain_order` option to use for the polars join.
+
+        Returns:
+            The collection, with members potentially reduced in length.
 
         Raises:
             ValueError: If the collection contains any member that is annotated with
                 `ignored_in_filters=True`.
 
-        Returns:
-            The collection, with members potentially reduced in length.
+        Attention:
+            This method does not validate the resulting collection. Ensure to only use
+            this if the resulting collection still satisfies the filters of the collection.
+            The joins are not evaluated eagerly. Therefore, a downstream call to `collect`
+            might fail, especially if `primary_keys` does not contain all columns for all
+            common primary keys.
         """
         if any(member.ignored_in_filters for member in self.members().values()):
             raise ValueError(
@@ -590,7 +599,12 @@ class Collection(BaseCollection, ABC):
 
         return self.cast(
             {
-                key: lf.join(primary_keys, on=self.common_primary_keys(), how=how)
+                key: lf.join(
+                    primary_keys,
+                    on=self.common_primary_keys(),
+                    how=how,
+                    maintain_order=maintain_order,
+                )
                 for key, lf in self.to_dict().items()
             }
         )
--- a/dataframely/collection.py
+++ b/dataframely/collection.py
@@ -587,10 +587,10 @@ class Collection(BaseCollection, ABC):
 
         Attention:
             This method does not validate the resulting collection. Ensure to only use
-            this if the resulting collection still satisfies the filters of the collection.
-            The joins are not evaluated eagerly. Therefore, a downstream call to `collect`
-            might fail, especially if `primary_keys` does not contain all columns for all
-            common primary keys.
+            this if the resulting collection still satisfies the filters of the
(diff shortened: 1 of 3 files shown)


dataframely/collection.py
```diff
@@ -561,6 +561,47 @@ def filter(
...
+        all_columns = next(iter(self.member_schemas().values())).columns()
+        # TODO think about ignored_in_filters
+        common_primary_keys = _common_primary_keys(self.member_schemas().values())
+        common_primary_key_columns = {
+            column: all_columns[column] for column in common_primary_keys
+        }
+        target_schema = create_schema(
+            name="CommonPrimaryKeysSchema", columns=common_primary_key_columns
+        )
+        validated_df = target_schema.validate(primary_keys, cast=True).lazy()
```
borchero, line 597: I'm not a big fan of this part here for two major reasons:

- IMO, no code from `testing` should be on a productive code path
- This currently eliminates laziness as `validate` requires collection

However, as we only allow semi- and anti-joins, _there's no possibility to mess up_ as we can never duplicate any rows (as in an inner-join). We don't even need to explicitly check whether the required primary key columns are provided in the input, `join` will take care of this :)

borchero (commented):
Thanks for introducing this! 😁

dataframely/collection.py
MoritzPotthoffQC: Simplified this now, thanks for the input!

dataframely/collection.py
```diff
@@ -561,6 +561,43 @@ def filter(
...
+        # TODO think about ignored_in_filters
+        # Get common primary keys, also considering schemas that are ignored in filters.
+        common_primary_keys = list(_common_primary_keys(self.member_schemas().values()))
+
+        return self.cast(
+            {
+                key: lf.join(primary_keys, on=common_primary_keys, how=how)
+                for key, lf in self.to_dict().items()
+            }
+        )
```
MoritzPotthoffQC, line 599: @borchero one thing I am not sure about: We currently ignore schemas that are annotated with `ignore_in_filters` in the default `common_primary_key()` method. Afaict, all joins across members of the data frame are done with that primary key. 

I am not sure if we should also ignore these primary keys here. Naively, I would assume that `join` also affects members that have `ignore_in_filters` set, but I have no good intuition as I never saw that option in use. 
In that case, I think we need the "full" common primary key here. 

If we do it this way, wording is a bit messy as all other usages of "common primary key" refer to something different and even the docs of `CollectionMember` mention that this annotation affects how the common primary key is defined.

dataframely/collection.py
```diff
@@ -561,6 +561,42 @@ def filter(
 
         return cls._init(results), failures
 
+    def join(
+        self,
+        primary_keys: pl.LazyFrame,
+        how: Literal["semi", "anti"] = "semi",
+    ) -> Self:
+        """Filter the collection by joining onto a data frame containing entries for the
+        common primary keys whose respective rows should be kept or removed in the
```
borchero, line 570:
```suggestion
        common primary key columns whose respective rows should be kept or removed in the
```

dataframely/collection.py
```diff
@@ -561,6 +561,42 @@ def filter(
...
+        collection members.
+
+        Args:
+            primary_keys: The data frame to join on. Must contain the common primary key
+                columns of the collection.
+            how: The join strategy to use. Like in polars, `semi` will keep all rows
+                that correspond to `primary_keys`, `anti` will remove them.
```
borchero, line 577:
```suggestion
                that can be found in `primary_keys`, `anti` will remove them.
```

dataframely/collection.py
```diff
@@ -561,6 +561,42 @@ def filter(
...
+
+        Raises:
+            ValueError: If the collection contains any member that is annotated with
+                `ignored_in_filters==True`.
```
borchero, line 581:
```suggestion
                `ignored_in_filters=True`.
```

dataframely/collection.py
```diff
@@ -561,6 +561,42 @@ def filter(
...
+            ColumnNotFoundError: If the `primary_keys` data frame does not contain
+                all common primary key columns of the collection.
```
borchero, line 583: This only happens upon collection, let's remove it
```suggestion
```

dataframely/collection.py
```diff
@@ -561,6 +561,42 @@ def filter(
 
         return cls._init(results), failures
 
+    def join(
+        self,
+        primary_keys: pl.LazyFrame,
+        how: Literal["semi", "anti"] = "semi",
```
borchero, line 567: Should we also add the `maintain_order` parameter? 
```suggestion
        how: Literal["semi", "anti"] = "semi",
        maintain_order: Literal["none", "left"] = "none",
```

dataframely/collection.py
```diff
@@ -561,6 +561,42 @@ def filter(
...
+    ) -> Self:
+        """Filter the collection by joining onto a data frame containing entries for the
```
borchero, line 569: I'm wondering if we still need an `Attention:` block that says that it's the user's responsibility that the resulting collection is still valid (just like in `cast`). You can probably construct custom rules that could be violated by running this join.

dataframely/collection.py
MoritzPotthoffQC: Yes makes sense 👍

borchero (approved):
Thank you!

Issue #130, opened 2025-08-25
Decimals with explicit precision and a min/max of 0 raise math domain error during validation


jhall-bp:
Hi,

When I try to create a schema (using v1.9.0) that contains a decimal column with an explicit precision like so:
 
```python
from decimal import Decimal
import dataframely as dy

_MIN = Decimal(0)

class MySchema(dy.Schema):
    dec = dy.Decimal(8, min=_MIN)
```

I get the following error:

```bash
Traceback (most recent call last):
  File "/Users/jhall-bp/Developer/test.py", line 8, in <module>
    class MySchema(dy.Schema):
        dec = dy.Decimal(8, min=Decimal())
  File "/Users/jhall-bp/Developer/test.py", line 9, in MySchema
    dec = dy.Decimal(8, min=Decimal())
  File "/Users/jhall-bp/Developer/.venv/lib/python3.13/site-packages/dataframely/columns/decimal.py", line 75, in __init__
    _validate(min, precision, scale, "min")
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jhall-bp/Developer/.venv/lib/python3.13/site-packages/dataframely/columns/decimal.py", line 172, in _validate
    if precision is not None and _num_digits(int(value)) > precision - scale:
                                 ~~~~~~~~~~~^^^^^^^^^^^^
  File "/Users/jhall-bp/Developer/.venv/lib/python3.13/site-packages/dataframely/columns/decimal.py", line 177, in _num_digits
    return int(math.log10(i) + 1)
               ~~~~~~~~~~^^^
ValueError: math domain error
```

It looks like this is caused by [this validation function here](https://github.com/Quantco/dataframely/blob/d8916e857180606e06a55893457cf173db73cb02/dataframely/columns/decimal.py#L172) where the min/max value is passed to the math.log10 function.

math.log10 doesn't like negative numbers either, so I think the _num_digits should look something like:

```python
def _num_digits(i: int) -> int:
    if i == 0:
        return 1
    return int(math.log10(abs(i)) + 1)
```

I also checked the python decimal documentation to see if there was already a way to do this but I couldn't find anything. It seems like you could also do the following which would also handle NaN and infinity though:

```python
def _num_digits(d: Decimal) -> int:
    (_sign, digits, exponent) = d.as_tuple()
    if not isinstance(exponent, int):  # ie if we have NaN or infinity
        return 0
    if exponent >= 0:
[...]

PR #131, opened 2025-08-29
fix: Correctly handle min and max bounds less than 1 for Decimal (#130)

jhall-bp:
# Motivation

Issue: #130 

`dy.Decimal` currently crashes during validation when instantiated with a min or max value that is 0 or less. This PR changes the way that the number of digits before the decimal place of a `decimal.Decimal` is calculated so that smaller mix/max values can be used as constraints.

# Changes

- The _num_digits helper function used for validating `dy.Decimal` inputs has been changed to use the digits and exponent values from `decimal.Decimal.as_tuple()` to calculate the number of digits instead of using `math.log10`
- Update test cases for `dy.Decimal`

### DIFF
--- a/dataframely/columns/decimal.py
+++ b/dataframely/columns/decimal.py
@@ -4,7 +4,6 @@
 from __future__ import annotations
 
 import decimal
-import math
 from typing import Any
 
 import polars as pl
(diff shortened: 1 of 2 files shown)


borchero (approved):
LGTM, thank you!

PR #134, opened 2025-09-02
feat: Add support for deltalake

AndreasAlbertQC:
# Motivation

I would like to be able to serialize data to deltalake, including the dataframely metadata.

# Changes

* Refactored the tests from `test_read_write_parquet.py` such that they can easily be run for each storage backend.
    * Since different backends use different args / kwargs,  support slightly different features (e.g. no truly lazy sink in deltalake atm), and need different code for mocking, I implemented a thin wrapper `Tester` interface that translates the calls needed for the tests into the specifics of the backends.
    * Tests are then simply `pytest.mark.parametrize`d over the differet backend testers
    * If we want to add another backend in the future, you would just implement the tester class and can then fully reuse existing tests. 
* Added implementation for reading from and writing to delta lake in Schema, Collection, and FailureInfo
    * data is stored in delta tables
    * meta data is stored in custom commit meta data. I initially considered table metadata, but switched to commits because table metadata is not changeable after an initial write, and may not reflect the current state of the table. With commit metadata, we can make sure to only trust the metadata if the last write came from dataframely. 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1211217826559584


dataframely/_storage/delta.py
```diff
@@ -0,0 +1,201 @@
...
+def _read_serialized_rules(
+    table: "deltalake.DeltaTable",
+) -> SerializedRules | None:
+    [last_commit] = table.history(limit=1)
+    return last_commit.get(RULE_METADATA_KEY, None)
+
+
+def _to_delta_table(
+    table: "Path | str | deltalake.DeltaTable",
+) -> "deltalake.DeltaTable":
```
borchero, line 192: You can just use `deltalake.DeltaTable` in the annotations (i.e. without quotes) here if you use `from __future__ import annotations`

dataframely/_storage/delta.py
```diff
@@ -0,0 +1,201 @@
...
+                target=uri / key,
+                **kwargs,
+            )
+
+    def scan_collection(
+        self, members: Iterable[str], **kwargs: Any
+    ) -> tuple[dict[str, pl.LazyFrame], list[SerializedCollection | None]]:
+        import deltalake
+
+        uri = Path(kwargs.pop("source"))
```
borchero, line 104: This could be an S3-path, which is not compatible with `Path`

borchero: We _could_ use `universal_pathlib` with `fsspec` but that's quite the dependency tree to add

dataframely/_storage/delta.py
```diff
@@ -0,0 +1,201 @@
...
+
+if TYPE_CHECKING:
+    import deltalake
+
+
+class DeltaStorageBackend(StorageBackend):
+    def sink_frame(
+        self, lf: pl.LazyFrame, serialized_schema: SerializedSchema, **kwargs: Any
+    ) -> None:
+        _raise_on_lazy_write()
```
borchero, line 27: Why the indirection with the separate method?

tests/collection/test_read_write_parquet.py
borchero, line 1: What's the purpose of this if we also have `test_storage.py`?

tests/schema/test_storage.py
```diff
@@ -0,0 +1,245 @@
...
+    tester.write_typed(wrong, df, tmp_path, lazy=lazy)
+    # Act
+    spy = mocker.spy(right, "validate")
+    tester.read(right, tmp_path, lazy, validation="skip")
+
+    # Assert
+    spy.assert_not_called()
+
+
+# ---------------------------- DELTA LAKE SPECIFICS ---------------------------------- #
```
borchero, line 238: [nit] This applies to everywhere: could you ensure empty newlines after these comments? Otherwise, they are interpreted to belong to the first method below :pp

borchero (commented):
Thank you, this looks great! The abstraction really shows its worth here :smile:

One additional request: can we add `deltalake` as an optional dependency to `pyproject.toml`?

dataframely/_storage/_exc.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) QuantCo 2025-2025
+# SPDX-License-Identifier: BSD-3-Clause
+
+
+def assert_metadata(metadata: str | None) -> str:
+    if metadata:
+        return metadata
+    raise ValueError(
+        "The required FailureInfo metadata was not found in the storage backend."
```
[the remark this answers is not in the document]
AndreasAlbertQC: It is indeed only used by the failure info serialization, but by both `parquet` and `delta` , so i need to import it from somewhere 🤔 What would be a better place?

dataframely/_storage/delta.py
AndreasAlbertQC: I raise the same exception a bunch of times (also on `sink_collection`, `sink_failure_info`), so I decided to factor it out

dataframely/_storage/delta.py
AndreasAlbertQC: Good point.

> that's quite the dependency tree to add

What do you mean by this? upath depends on fsspec, but fsspec has no required dependencies other than python, so it seems rather lightweight

dataframely/_storage/delta.py
AndreasAlbertQC: ahh so that's why I only needed the quotes sometimes and not others. 👍

dataframely/_storage/delta.py
```diff
@@ -0,0 +1,201 @@
...
+    SerializedCollection,
+    SerializedRules,
+    SerializedSchema,
+    StorageBackend,
+)
+from ._exc import assert_metadata
+from .constants import COLLECTION_METADATA_KEY, RULE_METADATA_KEY, SCHEMA_METADATA_KEY
+
+if TYPE_CHECKING:
+    import deltalake
```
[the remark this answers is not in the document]
AndreasAlbertQC: Good point, `_compat` is better, moved everything accordingly now.

dataframely/_storage/delta.py
```diff
@@ -0,0 +1,201 @@
...
+        return lf, serialized_rules, serialized_schema
+
+
+def _raise_on_lazy_write() -> None:
+    raise NotImplementedError("Lazy writes are not currently supported for deltalake.")
+
+
+def _read_serialized_schema(table: "deltalake.DeltaTable") -> SerializedSchema | None:
+    [last_commit] = table.history(limit=1)
+    return last_commit.get(SCHEMA_METADATA_KEY, None)
```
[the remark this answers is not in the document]
borchero: I think this is the responsibility ot the user. I would propose to document that clearly though in the docstrings for `read_delta` and `scan_delta`

dataframely/_storage/delta.py
borchero: Ah, I missed this, sorry; fine then!

AndreasAlbertQC:
Let's figure out the treatment of optional dependency testing in https://github.com/Quantco/dataframely/pull/135

PR #135, opened 2025-09-04
ci: Test optional dependencies separately

AndreasAlbertQC:
# Motivation

While working on #134, we noticed that we currently do not have stringent testing to ensure that our package works without optional dependencies.

# Changes

* Added `optional-dependencies` to `pyproject.toml`
* Refactored `pixi` environments such that we have separate testing environments with and without optional dependencies installed. The default environment still has the optional dependencies.
* Added a pytest marker `with_optionals`. This marker is used to mark all tests that require optional dependencies.
* In CI, we now use environments without optional dependencies to execute all tests that do not use them. A separate environment is used to run tests that require optionals to be installed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,9 @@ jobs:
           - os: ubuntu-latest
             environment: py313-optionals
             with_optionals: true
+          - os: windows-latest
+            environment: py313-optionals
+            with_optionals: true
     steps:
       - name: Checkout branch
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -102,11 +105,12 @@ install-polars-nightly = "pip install --pre --no-deps --upgrade --only-binary :a
...
-default = ["dev", "lint", "test"]
+default = ["dev", "lint", "test", "optionals"]
 docs = ["docs"]
 lint = { features = ["lint"], no-default-feature = true }
 nightly = ["nightly", "test"]
 py310 = ["py310", "test"]
 py311 = ["py311", "test"]
 py312 = ["py312", "test"]
 py313 = ["py313", "test"]
+py313-optionals = ["py313", "test", "optionals"]
```
AndreasAlbertQC, line 116: I considered running also the environments twice, but that felt like overkill to me.

borchero (approved):
Thanks, really useful! There was at least one time where we only noticed an issue in the conda-forge build 😅

PR #139, opened 2025-09-10
feat: Properly support S3 in IO functions

### LINKED ISSUE #87: Extend `write_parquet` and `read/scan_parquet` for `Collection` and `Schema` beyond filesystem locations
The current implementation is limited to (local) filesystem support (e.g., [here](https://github.com/Quantco/dataframely/blob/a3b90c2a1246309fcfb94cc561cf86e3bca35f93/dataframely/collection.py#L675-L678)). We should extend it to also support S3 locations or other storage backends.

borchero:
# Motivation

Fixes #87.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,7 +108,7 @@ filterwarnings = [
   "ignore::DeprecationWarning:botocore.*:",
 ]
 markers = [
-  "s3: tests that require access to S3-compatible storage",
+  "s3: tests that run against and S3 backend",
   "with_optionals: tests that require optional dependencies to be installed",
 ]
 testpaths = ["tests"]
--- a/dataframely/_storage/delta.py
+++ b/dataframely/_storage/delta.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import Any
 
 import polars as pl
+from fsspec import AbstractFileSystem, url_to_fs
 
 from dataframely._compat import deltalake
 
@@ -82,7 +83,8 @@ class DeltaStorageBackend(StorageBackend):
         serialized_schemas: dict[str, str],
         **kwargs: Any,
     ) -> None:
-        uri = Path(kwargs.pop("target"))
+        uri = kwargs.pop("target")
+        fs: AbstractFileSystem = url_to_fs(uri)[0]
 
         # The collection schema is serialized as part of the member parquet metadata
         kwargs["metadata"] = kwargs.get("metadata", {}) | {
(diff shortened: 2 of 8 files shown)


borchero:
Requires https://github.com/pola-rs/polars/issues/23346

docker-compose.yml
```diff
@@ -8,3 +7,15 @@ services:
       SA_PASSWORD: P@ssword1
     ports:
       - "1455:1433"
+
+  minio:
```
delsner, line 11: Did you consider using moto for s3 tests? It's easy to use and doesn't require running a container in the background (see http://docs.getmoto.org/en/latest/docs/server_mode.html#start-within-python).

borchero: cf. https://github.com/pola-rs/polars/pull/24443

pixi.toml
```diff
@@ -11,9 +11,9 @@ postinstall = "pip install --no-build-isolation --no-deps --disable-pip-version-
 python = ">=3.10"
 rust = "=1.85"
 
+fsspec = "*"
```
delsner:
```suggestion
fsspec = ">=2025.9"
```
they changed between semver and calver at some point.

pixi.toml
```diff
@@ -40,6 +40,8 @@ docs = "cd docs && make html"
 readthedocs = "rm -rf $READTHEDOCS_OUTPUT/html && cp -r docs/_build/html $READTHEDOCS_OUTPUT/html"
 
 [feature.test.dependencies]
+boto3 = "*"
```
delsner:
```suggestion
boto3 = ">=1.40.55"
```
they switched to `noarch` at some point.

pyproject.toml
```diff
@@ -16,8 +16,9 @@ classifiers = [
   "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
+  "fsspec",
```
delsner:
```suggestion
  "fsspec>=>=2025.9",
```

pyproject.toml
```diff
@@ -29,9 +30,10 @@ version = "0.0.0"
 
 [project.optional-dependencies]
 deltalake = ["deltalake"]
-sqlalchemy = ["sqlalchemy"]
 pyarrow = ["pyarrow"]
 pydantic = ["pydantic>=2"]
+s3 = ["s3fs"]
```
delsner:
```suggestion
s3 = ["s3fs>=2025.9"]
```

delsner (approved):
Looks good, just a few comments on the version pins.

tests/collection/test_parquet_s3.py
```diff
@@ -0,0 +1,82 @@
...
+
+class MyCollection(dy.Collection):
+    first: dy.LazyFrame[MyFirstSchema]
+    second: dy.LazyFrame[MySecondSchema] | None
+
+
+# --------------------------------------- TESTS -------------------------------------- #
+
+
+def test_read_write_parquet(s3_tmp_path: str) -> None:
```
AndreasAlbertQC: Have you considered to extend the testing abstractions `ParquetSchemaStorageTester`, `ParquetCollectionStorageTester`, `ParquetFailureInfoStorageTester` to also cover S3? (Or alternativey just implement separate ones like `S3ParquetSchemaStorageTester` etc)

The advantage of this approach would be that you get the entire coverage of the existing storage tests for free. With the current test, we are missing coverage for `FailureInfo`, as well as schema deserialization.

PR #144, opened 2025-09-11
fix: Use covariant Sequence and Mapping in check parameter

ytausch:
# Motivation

See `test_check_argument_covariant`. Mypy does not accept arguments for `Column.check` that do not exactly have the correct type:

```text
error: Argument "check" to "String" has incompatible type "dict[str, Callable[[Expr], Expr]]"; expected "Callable[[Expr], Expr] | list[Callable[[Expr], Expr]] | dict[str, Callable[[Expr], Expr]] | None"  [arg-type]
```

While this particular error looks like a mypy bug since `dict[str, Callable[[Expr], Expr]]` is a subtype of `dict[str, Callable[[Expr], Expr]]` unioned with other things, I found that it might be good practice anyway to accept the covariant types `Sequence` and `Mapping` instead of the invariant types `list` and `dict`.

# Changes

Changed the `Check` type alias to use `Sequence` and `Mapping` instead of `list` and `dict`.

All that said, using the covariant types resolves all issues mypy has with the test I've added. See also [these Python 3.11 docs](https://docs.python.org/3.11/library/typing.html#typing.Dict) (not sure where it went for newer Python versions):

> Note that to annotate arguments, it is preferred to use an abstract collection type such as [Mapping](https://docs.python.org/3.11/library/typing.html#typing.Mapping) rather than to use [dict](https://docs.python.org/3.11/library/stdtypes.html#dict) or typing.Dict.

### DIFF
--- a/dataframely/columns/_base.py
+++ b/dataframely/columns/_base.py
@@ -145,18 +145,20 @@ class Column(ABC):
                 for rule_name, rule_callable in self.check.items():
                     result[f"check__{rule_name}"] = rule_callable(expr)
             else:
-                rules_sequence = (
-                    self.check if isinstance(self.check, Sequence) else [self.check]
+                list_of_rules = (
+                    list(self.check)
+                    if isinstance(self.check, Sequence)
+                    else [self.check]
                 )
                 # Get unique names for rules from callables
-                rule_names = self._derive_check_rule_names(rules_sequence)
-                for rule_name, rule_callable in zip(rule_names, rules_sequence):
+                rule_names = self._derive_check_rule_names(list_of_rules)
+                for rule_name, rule_callable in zip(rule_names, list_of_rules):
                     result[rule_name] = rule_callable(expr)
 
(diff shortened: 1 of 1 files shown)


tests/columns/test_check.py
```diff
@@ -12,6 +12,22 @@ class CheckSchema(dy.Schema):
...
+        return (expr == "all").all() | (expr != "all").all()
+
+    check_dict = {"all_or_none": check_all_or_none}
+
+    class CheckDictSchema(dy.Schema):
+        column = dy.String(check=check_dict)
+
+    df = pl.DataFrame({"column": ["all", "all", "all"]})
+    _, failures = CheckDictSchema.filter(df)
+    assert failures.counts() == {}
```
borchero, line 28: For future reference: there is also `test_typing.py` whose purpose is to make sure that `mypy` passes -- the tests are not actually run in that file :smile:

PR #147, opened 2025-09-16
fix: Fix serialization of `dy.Enum` columns

MoritzPotthoffQC:
# Motivation

The possible values are now stored as `pl.Series` (#138), which can't be json-serialized. 

# Changes

Convert the values to a python list before serialization.


borchero (commented):
This is the second issue that we've run into after using `Series` instead of `list` here, let's please store the categories as `list` 😅 fyi @AndreasAlbertQC

PR #158, opened 2025-10-06
fix: Fix `max_exclusive=0` being ignored

AndreasAlbertQC:
# Motivation

We do not properly differentiate between `max_exclusive=None` and `max_exclusive=0` because we naively cast to bool.

# Changes

* Added a test that catches the bug
* Fixed the bug


borchero (approved):
Whoops, thank you!

Issue #126, opened 2025-08-14
Add warning or fail validation when schema definition is invalid


denskh:
Invalid schema definition leads to silently dropped columns in validated dataframe.

Repro sample:

```python
import dataframely as dy
import polars as pl

class HouseSchema(dy.Schema):
    num_bedrooms = dy.UInt8(nullable=True)
    price = dy.Float64  **# USER ERROR**

df = pl.DataFrame({
    "num_bedrooms": [2, 2, 1, None, None, 2],
    "price": [100_000, 110_000, 50_000, 80_000, 60_000, 160_000]
})

df2: dy.DataFrame[HouseSchema] = HouseSchema.validate(df, cast=True)
print(set(df.columns) - set(df2.columns))
```

Output:
`{'price'}`

Ideal behaviors in my opinion would be failed validation or at least warning.

Issue #105, opened 2025-07-30
Warn on silent drop of tuple of Columns


csanders-census:
If you accidently pass in a tuple of columns:

```
class MySchema(dy.Schema):
    var1 = dy.Int32(nullable=False, primary_key=True)
    var2 = dy.Int32(nullable=False),
    var3 = dy.Int32(nullable=False, min=1, max=100)
```

the resulting validated frame will only contain var1 and var3 with no warning issued. Static type checkers also don't see a problem. 

This is not a bug as it's certainly user error, but I just hit my head against the wall for half an hour on this, so even this issue may help future users. I don't know what the best solution is here; changing SchemaMeta._get_metadata to 
```
if isinstance(value, Columns):
    ...
elif isinstance(value, Rule):
    ...
else:
   raise ValueError("Class members must be a column or a rule.")
```
seems like an easy fix if it's consistent with the rest of the API.

delsner:
Thanks for reporting. As you say this is a user error and I would recommend you to give [`black`](https://github.com/psf/black) a try which wraps `var2 = dy.Int32(nullable=False),` with `(...)` leading to `var2 = (dy.Int32(nullable=False),)` which is easier to spot.

In general, I'm open to adding `raise ValueError("Class members must be a column or a rule.")` to `SchemaMeta._get_metadata` if that's really only what it needs. Wdyt @borchero @AndreasAlbertQC?

csanders-census:
I do use ruff but simply hadn't run it yet! Life lesson that linters are valuable debugging tools :)

AndreasAlbertQC:
@delsner when we want to extend the list of allowed types, we can just also adjust the check, right? Feels OK to me to explicitly check this and throw an error if we get something unexpected. Only nitpicky opinion I have is that this should probably be a `TypeError` instead of a `ValueError` :)

AndreasAlbertQC:
@jjurm you have a point: we currently allow users to have extra members and simply disallowing this going forward would be breaking change. 

> allow opt-out: have a configurable class attribute extra_members: Literal["forbid", "allow"] = "forbid" that can be set to "allow" if needed (similar to pydantic)

I think configuration might come sooner or later, #100 also already touched on it. I think we should start collecting things we want to configure and plan what configuration mechanism we want to use. I'd prefer not to have to solve that right here right now.

> just issue a UserWarning

[...]

PR #159, opened 2025-10-06
feat: Add type error for pathological non-`Column` members in `Schema`

aaronweissberg-lab:
…es of dy column and types instead of instance caused by user errors)

# Motivation

As described in issues #105 and #126 columns are silently dropped if by user error one adds a trailing comma (creating a tuple) or forgets brackets (column type instead of instance)

# Changes

Tests for non-column members as described were added to test_base.py. SchemaMeta was updated to raise a TypeError under the specified conditions (could be changed to warning if preferable).

### DIFF
--- a/dataframely/_base_schema.py
+++ b/dataframely/_base_schema.py
@@ -156,7 +156,7 @@ class SchemaMeta(ABCMeta):
             if isinstance(value, type) and issubclass(value, Column):
                 raise TypeError(
                     f"Column '{attr}' is a type, not an instance. "
-                    f"Schema members must be of type Column not type[Column]."
+                    f"Schema members must be of type Column not type[Column]. "
                     f"Did you forget to add parentheses?"
                 )
 


dataframely/_base_schema.py
```diff
@@ -137,6 +137,25 @@ def __new__(
                         f"which are not in the schema: {missing_list}."
                     )
 
+        # 3) Check that all members are non-pathological (i.e., user errors).
+        for attr_name, attr_value in namespace.items():
+            if attr_name.startswith("_"):
+                continue  # skip internal names
```
jjurm, line 143: This condition seems too wide, I intendedly use the `_` prefix to distinguish some columns

AndreasAlbertQC, line 143:
```suggestion
                continue
```

dataframely/_base_schema.py
```diff
@@ -137,6 +137,25 @@ def __new__(
...
+
+            # Check for tuple of column (commonly caused by trailing comma)
+            if isinstance(attr_value, tuple):
```
jjurm, line 146: What about raising this only when the tuple has length 1? (to be on the conservative side)

dataframely/_base_schema.py
```diff
@@ -137,6 +137,25 @@ def __new__(
...
+                raise TypeError(
+                    f"Column '{attr_name}' is defined as a tuple. "
+                    f"Did you accidentally add a trailing comma?"
+                )
+
+            # Check for column type instead of instance (e.g., dy.Float64 instead of dy.Float64())
+            if isinstance(attr_value, type) and issubclass(attr_value, Column):
+                raise TypeError(
+                    f"Column '{attr_name}' is a type, not an instance. "
+                    f"Did you forget to add parentheses?"
```
jjurm, line 156: I'd suggest the message be a little bit more explicit in mentioning something along the lines of "schema members must be of type `Column`, not `type[Column]`"

jjurm (commented):
Thanks for the contribution @aaronweissberg-lab ! 🚀

dataframely/_base_schema.py
AndreasAlbertQC, line 146:
```suggestion
            if (
                isinstance(attr_value, tuple)
                and len(attr_value) > 0 
                and isinstance(attr_value[0], dy.Column)
             ):
```

I think we want to specifically target cases where the attribute is a tuple of columns, to not forbid users from having tuples in general.

dataframely/_base_schema.py
```diff
@@ -137,6 +137,25 @@ def __new__(
...
 
+        # 3) Check that all members are non-pathological (i.e., user errors).
+        for attr_name, attr_value in namespace.items():
+            if attr_name.startswith("_"):
+                continue  # skip internal names
+
+            # Check for tuple of column (commonly caused by trailing comma)
+            if isinstance(attr_value, tuple):
+                raise TypeError(
+                    f"Column '{attr_name}' is defined as a tuple. "
```
AndreasAlbertQC, line 148:
```suggestion
                    f"Column '{attr_name}' is defined as a tuple of dy.Column. "
```

tests/schema/test_base.py
```diff
@@ -95,3 +95,20 @@ def test_dunder_name() -> None:
 
 def test_dunder_name_with_rule() -> None:
     assert MySchemaWithRule.__name__ == "MySchemaWithRule"
+
+
+def test_user_error_tuple_column() -> None:
```
AndreasAlbertQC, line 100: I think we are currently still missing a test that shows that the user is allowed to specify extra members (i.e. having a valid, non-column member in a Schema). Could you add one?

AndreasAlbertQC (commented):
Thanks @aaronweissberg-lab ! I think this is great and I have only minor suggestions :)

dataframely/_base_schema.py
AndreasAlbertQC: This should probably made consistent with [this line](https://github.com/Quantco/dataframely/blob/main/dataframely/_base_schema.py#L161), where we skip if theres a double underscore

tests/schema/test_base.py
AndreasAlbertQC: probably a tuple of something other than `dy.Column` so that we know that our check above works correctly

dataframely/_base_schema.py
```diff
@@ -137,6 +137,29 @@ def __new__(
                         f"which are not in the schema: {missing_list}."
                     )
 
+        # 3) Check that all members are non-pathological (i.e., user errors).
+        for attr, value in {
+            k: v for k, v in namespace.items() if not k.startswith("__")
+        }.items():
```
AndreasAlbertQC, line 143:
```suggestion
        for attr, value in namespace.items():
            if attr.startswith("__"):
                continue
```

I'd prefer this because 1. I find it easier to read and 2. it avoids an unnecessary second loop pass

AndreasAlbertQC: I'm just gonna commit this change :)

dataframely/_base_schema.py
```diff
@@ -137,6 +137,29 @@ def __new__(
                         f"which are not in the schema: {missing_list}."
                     )
 
+        # 3) Check that all members are non-pathological (i.e., user errors).
```
borchero, line 140: One question: why are we asserting that types are specific disallowed types instead of checking that they are _allowed_. It seems like we'd be missing other invalid types here 🤔

AndreasAlbertQC: Before this PR, we have already been allowing users to have arbitrary extra members. To not break this, we decided to just raise explicitly for stuff where we are 100% sure that the user cannot have intended what they did.

Issue #154, opened 2025-10-01
Request: Add a way (with docs examples) to set multiple primary/unique keys


DeflateAwning:
Oftentimes, due to database migrations, etc., there are multiple primary keys, etc. that we want to track. Is there a way to add that to the schema validation. 

For example, I have `column_a`, `column_b`, and `column_a_b_concat`. I want to assert `column_a_b_concat` as the main primary key, and also `["column_a", "column_b"]` as another primary key.

PR #163, opened 2025-10-06
docs: Add multiple unique keys FAQ recipe

DeflateAwning:
# Motivation

Related to #154 - feature request turned into a documentation request.

# Changes

Adding the first FAQ!

### DIFF
--- a/docs/sites/faq.rst
+++ b/docs/sites/faq.rst
@@ -7,7 +7,7 @@ thinking, please add it here.
 How do I define additional unique keys in a ``dy.Schema``?
 ----------------------------------------------------------
 
-By default, ``dataframely`` only supports defining a single non-nullable primary key in ``dy.Schema``.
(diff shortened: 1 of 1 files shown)


docs/sites/faq.rst
```diff
@@ -3,3 +3,28 @@ FAQ
 
 Whenever you find out something that you were surprised by or needed some non-trivial
 thinking, please add it here.
+
+How do I define additional unique keys in a ``dy.Schema``?
+----------------------------------------------------------
+
+By default, ``dataframely`` only supports defining a single non-nullable primary key in ``dy.Schema``.
```
borchero, line 10: Since nomenclature around primary keys is slightly fuzzy, I want to make it clear here that dataframely _does_ support composite primary keys. How about adjusting this slightly:

```suggestion
By default, ``dataframely`` only supports defining a single non-nullable (composite) primary key in ``dy.Schema``.
```

borchero (commented):
Thanks for adding this immediately @DeflateAwning! :)

PR #174, opened 2025-10-17
feat!: Make columns non-nullable by default

AndreasAlbertQC:
# Motivation

Implements #20

# Changes

* Changed the default of `nullable` from `True` to `False`
* Replaced `FutureWarning`
* Added a test for the `skip_if` decorator we otherwise don't use anymore right now. I expect we will use it again for future deprecations, so I figured it should stay in the code base.


tests/column_types/test_datetime.py
```diff
@@ -216,47 +216,47 @@ def test_args_resolution_valid(
     ("column", "values", "valid"),
     [
         (
-            dy.Date(min=dt.date(2020, 4, 1)),
+            dy.Date(min=dt.date(2020, 4, 1), nullable=True),
```
borchero: Just a Q: Why do these have to be nullable?

tests/column_types/test_decimal.py
```diff
@@ -89,7 +89,10 @@ def test_non_decimal_dtype_fails(dtype: DataTypeClass) -> None:
     ],
 )
 def test_validate_min(inclusive: bool, valid: dict[str, list[bool]]) -> None:
-    kwargs = {("min" if inclusive else "min_exclusive"): decimal.Decimal(3)}
+    kwargs = {
+        ("min" if inclusive else "min_exclusive"): decimal.Decimal(3),
+        "nullable": True,
```
borchero: Again, why nullable?

tests/column_types/test_string.py
```diff
@@ -9,23 +9,23 @@
 
 
 def test_validate_min_length() -> None:
-    column = dy.String(min_length=2)
+    column = dy.String(min_length=2, nullable=True)
```
borchero: I think I don't understand some of these changes -- we never have NULLs in the data, do we?

borchero (commented):
Neat, just small questions related to the tests

tests/columns/test_pyarrow.py
```diff
@@ -17,9 +19,18 @@
 pytestmark = pytest.mark.with_optionals
 
 
+T = TypeVar("T", bound=dy.Column)
+
+
+def _nullable(column_type: type[T]) -> T:
+    if column_type == dy.Any:
+        return column_type()
```
delsner:
```suggestion
    # dy.Any doesn't have the `nullable` parameter.
    if column_type == dy.Any:
        return column_type()
```

PR #176, opened 2025-10-17
feat: Support Python 3.14

borchero:
# Motivation

Python 3.14 has been released on Oct 7 -- dataframely collections currently does not work with it.


dataframely/_base_collection.py
```diff
@@ -188,11 +191,17 @@ def _get_metadata(source: dict[str, Any]) -> Metadata:
...
         # Get all members via the annotations
-        if "__annotations__" in source:
-            for attr, kls in source["__annotations__"].items():
-                result.members[attr] = CollectionMeta._derive_member_info(
-                    attr, kls, CollectionMember()
-                )
+        annotations = {}
+        if sys.version_info >= (3, 14):
+            if "__annotate_func__" in source:
+                annotations = source["__annotate_func__"](Format.VALUE)
```
borchero: I'm not sure if this is worse than before but it feels a little whacky. I didn't figure out how to use `annotationlib.get_annotations` here though, I don't think that works within the metaclass. Maybe there's something to re-architecture at some point...

pixi.toml
```diff
@@ -103,17 +103,21 @@ python = "3.12.*"
...
 lint = { features = ["lint"], no-default-feature = true }
-nightly = ["nightly", "test", "optionals"]
+nightly = ["nightly", "optionals", "test"]
 py310 = ["py310", "test"]
 py311 = ["py311", "test"]
 py312 = ["py312", "test"]
 py313 = ["py313", "test"]
-py313-optionals = ["py313", "test", "optionals"]
+py314 = ["py314", "test"]
+py314-optionals = ["optionals", "py313", "test"]
```
delsner:
```suggestion
py314-optionals = ["optionals", "py314", "test"]
```
?

PR #179, opened 2025-10-20
feat: Introduce lazy validation

### LINKED ISSUE #93: Support fully lazy validation
# Motivation

Currently, `.validate` and `.filter` always require collecting the input data frame. This has two major drawbacks:

- When using dataframely as part of a larger data pipeline, it blocks global optimizations 
- It prevents proper use of the streaming engine

Ideally, `.validate` and `.filter` for both schemas and collections would behave as "lazy-in-lazy-out" ("eager-in-eager-out").

As an example: for the schema, this would result in the following signature for the `.validate` method.

```python
@overload
@classmethod
def validate(
    cls, df: pl.DataFrame, /, *, cast: bool = False
) -> DataFrame[Self]: ...

@overload
@classmethod
def validate(
    cls, df: pl.LazyFrame, /, *, cast: bool = False
) -> LazyFrame[Self]: ...

@classmethod
def validate(
    cls, df: pl.DataFrame | pl.LazyFrame, /, *, cast: bool = False
) -> DataFrame[Self] | LazyFrame[Self]: ...
```

borchero:
# Motivation

Fixes #93.

# Changes

- Introduce a custom polars plugin to allow for lazily raising on validation errors
- Add an `eager` parameter to `{Schema,Collection}.{validate,filter}` to allow for lazy validation
  - The default remains `True` as validation should, by default, serve as a barrier
- Fix the PyPI build to actually perform a release build (lol)

NOTE: The eager validation/filter performance is equivalent (and tends to be slightly faster than before). There are, however, significant performance penalties for lazy validation/filtering in collections as well as a notable performance drop for lazy filtering of schemas. This is likely largely due to https://github.com/pola-rs/polars/pull/24129.

### DIFF
--- a/dataframely/_base_schema.py
+++ b/dataframely/_base_schema.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 import sys
 import textwrap
-from abc import ABCMeta, abstractmethod
+from abc import ABCMeta
 from copy import copy
 from dataclasses import dataclass, field
 from typing import Any


tests/schema/test_validate.py
```diff
@@ -147,8 +142,8 @@ def test_group_rule_on_nulls(
 ) -> None:
     # The schema is violated because we have multiple "b" values for the same "a" value
     df = df_type({"a": [None, None], "b": [1, 2]})
-    with pytest.raises(RuleValidationError):
-        schema.validate(df, cast=True)
+    with pytest.raises(plexc.ComputeError, match=r"1 rules failed validation"):
```
AndreasAlbertQC: Does this have to be a generic ComputeError? I saw `validation_error.rs`, which seemed like there would still be a distinct `RuleValidationError`

dataframely/collection/collection.py
```diff
@@ -370,29 +372,86 @@ def validate(cls, data: Mapping[str, FrameType], /, *, cast: bool = False) -> Se
...
-        return out
+                details = [
+                    f" > Member '{member}' failed validation:\n"
+                    + textwrap.indent(error, "   ")
+                    for member, error in errors.items()
+                ]
+                message = "\n".join(
+                    [f"{len(errors)} members failed validation:"] + details
+                )
+                raise plexc.ComputeError(message)
```
AndreasAlbertQC: I assume you chose `ComputeError` here in order to be consistent with the lazy case. Is that right? While consistency is good, I feel sad that I don't get a nice `ValidationError` type here. How bad would it be to break consistency and raise a `ValidationError` in this branch? Consistency would mainly be useful for users who can use a single code path for both lazy and eager validation, but is that even a possible outcome? After all, even if it is the same exception type, I would have to deal with it in two different places (either when I call `validate` or when I call `collect`), which could be far apart. The more I think about it, the more I think that consistency is not really a concern here

tests/schema/test_validate.py
```diff
@@ -2,12 +2,13 @@
 # SPDX-License-Identifier: BSD-3-Clause
```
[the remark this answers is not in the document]
borchero: Fair, they should, I think I was not 100% sure what I did already and coverage didn't point out an issue 😅

borchero:
> Could take this opportunity to add a validation.md under docs/sites/features ?

This is still missing, I think this can be reviewed regardless :smile:

dataframely/_base_schema.py
```diff
@@ -23,7 +23,7 @@
 
 
 if TYPE_CHECKING:
-    from ._typing import DataFrame
+    pass
```
delsner: do we still need this?

dataframely/exc.py
```diff
@@ -1,91 +1,19 @@
...
-            )
-        ]
-        return "\n".join([f"{self.message}:"] + schema_details + column_details)
+class SchemaError(Exception):
+    """Error raised when the data frame schema does not match the dataframely schema."""
 
 
-class MemberValidationError(ValidationError):
-    """Validation error raised when multiple members of a collection fail validation."""
+class ValidationError(Exception):
```
delsner: nit: maybe call this `EagerValidationError` to clarify that lazy validation will not raise this error?

dataframely/collection/collection.py
```diff
@@ -370,29 +372,86 @@ def validate(cls, data: Mapping[str, FrameType], /, *, cast: bool = False) -> Se
                 the member as key.
             cast: Whether columns with a wrong data type in the member data frame are
                 cast to their schemas' defined data types if possible.
+            eager: Whether the validation should be performed eagerly. If ``True``, this
+                method raises a validation error and the returned collection contains
+                "shallow" lazy frames, i.e. lazy frames by simply calling
```
delsner:
```suggestion
                "shallow" lazy frames, i.e., lazy frames by simply calling
```

dataframely/collection/collection.py
```diff
@@ -370,29 +372,86 @@ def validate(cls, data: Mapping[str, FrameType], /, *, cast: bool = False) -> Se
...
+                :meth:`~polars.DataFrame.lazy` on the validated data frame. If
+                ``False``, this method only raises if ``data`` does not contain data for
```
delsner: iiuc a `ValueError` would be raised in this case, right?

dataframely/schema.py
```diff
@@ -332,7 +336,7 @@ def sample(
                 validation_error = None
                 try:
                     cls.validate(relevant_rows)
-                except ValidationError as e:
+                except Exception as e:
```
delsner: Can we further narrow this down? `ComputeError | ...`?

dataframely/schema.py
```diff
@@ -446,71 +450,192 @@ def _sampling_overrides(cls) -> dict[str, pl.Expr]:
...
+    @overload
+    @classmethod
+    def validate(
+        cls,
+        df: pl.DataFrame | pl.LazyFrame,
+        /,
+        *,
+        cast: bool = False,
+        eager: bool,
+    ) -> DataFrame[Self] | LazyFrame[Self]: ...
```
delsner: why do we need this overload given that `eager: bool` always has a default `True` (i.e., cannot be unset)?

pyproject.toml
```diff
@@ -41,8 +41,7 @@ Documentation = "https://dataframely.readthedocs.io/"
 Repository = "https://github.com/quantco/dataframely"
 
 [tool.maturin]
-module-name = "dataframely._extre"
-profile = "release"
```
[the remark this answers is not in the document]
borchero: Yep! Otherwise, every `maturin develop` performs a release build which is rather annoying. `pixi run postinstall` performs a release build regardless.

dataframely/filter_result.py
```diff
@@ -25,6 +32,56 @@
...
+    failure: FailureInfo[S]
+
+
+class LazyFilterResult(NamedTuple, Generic[S]):
+    """Container for results of calling :meth:`Schema.filter` on a lazy frame."""
+
+    #: The rows that passed validation.
+    result: LazyFrame[S]
+    #: Information about the rows that failed validation.
+    failure: FailureInfo[S]
```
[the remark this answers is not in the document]
borchero: I would generally like to keep it; after all, you're not forced to specify the generic parameter. Happy to discuss again in the future though.

dataframely/schema.py
borchero: Absolutely, this was surely a placeholder 😅

src/polars_plugin/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+    let result = match inputs.len() {
+        0 => polars_bail!(ComputeError: "cannot combine zero rules"),
+        1 => as_bool(&inputs[0])?.clone(),
+        2 => as_bool(&inputs[0])? & as_bool(&inputs[1])?,
+        n if n < POOL.current_num_threads() * 2 => inputs
+            .iter()
+            .skip(2)
+            .try_fold(as_bool(&inputs[0])? & as_bool(&inputs[1])?, |acc, s| {
+                as_bool(s).map(|b| &acc & b)
+            })?,
```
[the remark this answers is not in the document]
borchero: Yeah, I benchmarked this at some point, this seems to be the point where parallelism yields benefits.

PR #180, opened 2025-10-20
docs: Clean up small issues in docs

AndreasAlbertQC:
# Motivation

Polishing.

# Changes

* Added intersphinx support so we can link to other projects properly.
* Wherever applicable, I fixed up class/method/function references to render as hyperlinks
* Added the `napoleon` extension so our doc strings are parsed more robustly. This resolves small styling problems with lists of arguments in docstrings, as well as with Exceptions in docstrings
* Removed outdated docstring references to the mypy plugin

Note: you cannot currently link to `polars.DataFrame`  and `polars.LazyFrame`, which seems to be a problem in the `polars` doc build, not on our end. You can verify that these classes do not appear in the structured output sphinx uses: `pixi run -e docs python -m sphinx.ext.intersphinx https://docs.pola.rs/py-polars/html/objects.inv`

### DIFF
--- a/dataframely/schema.py
+++ b/dataframely/schema.py
@@ -633,7 +633,7 @@ class Schema(BaseSchema, ABC):
 
         Note:
             If you only require a generic data frame for the type checker, consider
-            using :func:`python:typing.cast` instead of this method.
+            using :func:`typing.cast` instead of this method.
 
         Attention:
             For lazy frames, casting is not performed eagerly. This prevents collecting


dataframely/collection.py
```diff
@@ -75,10 +75,6 @@ class MyCollection(dy.Collection):
     Besides, it may define *filters* (c.f. :meth:`~dataframely.filter`) and arbitrary
     methods.
 
-    Note:
```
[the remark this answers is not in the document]
borchero: You're right

dataframely/collection.py
```diff
@@ -962,14 +965,16 @@ def scan_delta(
                   data, entrusting the user that the schema is valid. *Use this option
                   carefully*.
 
-            kwargs: Additional keyword arguments passed directly to :meth:`polars.scan_delta`.
+            kwargs: Additional keyword arguments passed to :func:`polars.scan_delta`.
```
borchero: When do I use `:meth:`, when do I use `:func:`?

docs/sites/faq.md
```diff
@@ -3,9 +3,10 @@
 Whenever you find out something that you were surprised by or needed some non-trivial
 thinking, please add it here.
 
-## How do I define additional unique keys in a `dy.Schema`?
+## How do I define additional unique keys in a :class:`~dataframely.Schema`?
```
borchero: Why does this look different to below?
```suggestion
## How do I define additional unique keys in a {class}`~dataframely.Schema`?
```

borchero (commented):
Thanks a lot! Regarding polishing documentation: can we also get rid of the double backticks in the docstrings? :smile:

docs/sites/faq.md
AndreasAlbertQC: bug

dataframely/schema.py
```diff
@@ -629,11 +633,11 @@ def cast(
 
         Note:
             If you only require a generic data frame for the type checker, consider
-            using :meth:`typing.cast` instead of this method.
+            using :func:`python:typing.cast` instead of this method.
```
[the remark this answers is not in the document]
AndreasAlbertQC: nah

dataframely/collection.py
borchero: I always thought `:meth:` works for both. Does it not?

PR #188, opened 2025-10-23
docs: Add example for `SchemaClass.filter()` to docstring

DeflateAwning:
# Motivation

Yesterday, I was bothered when I couldn't find this and had to re-read the quickstart guide.

Today, I still couldn't remember how to do it. There are no hints once you have the `bad` dataclass instance. Thus, we show the example here.

# Changes
Added example to docstring. Copied out of quickstart guide. Having this example in two places makes a lot of sense.

### DIFF
--- a/dataframely/collection/collection.py
+++ b/dataframely/collection/collection.py
@@ -521,26 +521,6 @@ class Collection(BaseCollection, ABC):
                 Note that until https://github.com/pola-rs/polars/pull/24129 is
                 released, eagerly filtering can provide significant speedups.
 
-
-        Example:
-
-        .. code-block:: python
-
-            # Define collection
-            class HospitalInvoiceData(dy.Collection):
-                invoice: dy.LazyFrame[InvoiceSchema]
-                ...
-
-            # Filter the data and cast columns to expected types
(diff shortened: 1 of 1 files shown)


DeflateAwning:
Seems reasonable, but I'm not sure what that example should look like.

Can you suggest what the example should look like? Haven't used Collections yet, and the docs don't really have complete enough examples to copy from imo.

delsner:
> Can you suggest what the example should look like? Haven't used Collections yet, and the docs don't really have complete enough examples to copy from imo.

The example can be very similar to the schema example:

```python
# Define collection
class HospitalInvoiceData(dy.Collection):
    invoice: dy.LazyFrame[InvoiceSchema]
    ...

# Filter the data and cast columns to expected types
good, failure = HospitalInvoiceData.filter(df, cast=True)

# Inspect the reasons for the failed rows for member `invoice`
print(failure.invoice.counts())

# Inspect the failed rows
failed_df = failure.invoice.invalid()
print(failed_df)
```

DeflateAwning:
Review comments implemented. Rebased. Thanks.

delsner (approved):

PR #198, opened 2025-10-24
refactor!: Make `@dy.rule` decorator apply to classmethods

borchero:
# Motivation

https://github.com/Quantco/dataframely/discussions/185 -- there's no good reason not to support this properly.

# Changes

- Make `@dy.rule()` apply to classmethods, i.e. methods that receive `cls`
  - This makes the `mypy` plugin superfluous (nice!) but requires setting
    ```toml
    [tool.ruff.lint.pep8-naming]
    classmethod-decorators = ["dataframely.rule"]
    ```
  - We _could_ get rid of this by simply naming the receiver `self` but this feels wrong (since the method is never actually called on an instance). Please challenge if you think differently though.

Apart from this change, all other methods are backwards-compatible.


delsner (approved):
Thanks! I think the ruff rule is fine. Should we document this a bit more prominent than in the migration docs? (Sorry, if I missed it.) Also, we could think about changing some examples to use `cls.<column>.col` now. Wdyt?

PR #263, opened 2026-01-23
feat: Allow int values for Decimal type constraints

ytausch:
# Motivation

Being able to use `int`s directly in constraint definitions is a bit more ergonomic, and doesn't have any drawbacks as far as I see. 

# Changes

Changed the constraint definitions of `Decimal` to accept `int` values.


borchero (approved):
Makes sense, thanks!

Issue #122, opened 2025-08-12
Feature Request: Enhanced Soft Failure support


raayu83:
Just started experimenting with dataframely and liking a lot of what I see.

It would be great if Soft Failure support was extended in a way that makes it easy to generate failure reports for users who supply the input data.

Some ideas that sound useful to me:
- Provide a way to mark a set of columns as debug ID columns for Failure reports
- Extend FailureInfo counts and cooccurrence_counts methods with a "group_by_ids=True" or "group_by='id'" or similar parameter - if set, errors are grouped by the id columns suggested above
- it would be great if friendly names (that have a meaning to non technical users) could assigned to dy.rule

PR #274, opened 2026-02-16
feat: Allow users to set custom rule names

AndreasAlbertQC:
# Motivation

Addresses part of #122

# Changes

* Added `name` argument to `dy.rule`, which allows users to set a custom name that will show up in any failure outputs.


borchero (commented):
I'm not sure you can answer that question because you didn't write the initial issue report... :smile: but what's the rationale of assigning custom names to rules but not being able to provide custom names to column-level checks 🤔

Issue #272, opened 2026-02-08
Get list of validation results for each rule/check?


reinholdsson:
Hi,

Would it be possible to add support for returning (optionally) a dictionary or similar structure that maps each rule/check to its ValidationResult instance?

We have a use case for integrating dataframely's schema validation with Dagster asset checks. To enable this, we need a way to programmatically access each rule's validation result so we can dynamically map them to corresponding asset checks.

Currently, the validation output may look like:
```
ValidationError: 4 rules failed validation:
 - 'reasonable_bathroom_to_bedrooom_ratio' failed for 1 rows
 - 'minimum_zip_code_count' failed for 2 rows
 * Column 'zip_code' failed validation for 1 rules:
   - 'min_length' failed for 1 rows
 * Column 'num_bedrooms' failed validation for 1 rules:
   - 'nullability' failed for 2 rows
```

Hence, it would be great to have an API to obtain validation results in a structured mapping, e.g.:
```python
{
    "reasonable_bathroom_to_bedrooom_ratio": ValidationResult(...),
    "minimum_zip_code_count": ValidationResult(...),
    "zip_code_min_length": ValidationResult(...),
    ...,
}
```

This would allow us to:
- Dynamically map each check to a Dagster asset check.
- Retrieve per-rule validation details programmatically, not just from the textual summary.

Thanks! :)

borchero:
We should definitely improve the usability of `FailureInfo`; there's been a bunch of questions/requests around this (see also #122, #124, #190). I think all of these issues should give us enough of an idea what common use cases look like and we should extend the public API accordingly.

danielcrane:
I also feel like exposing a method that returns a dictionary with the validation results (incl. rows that failed the check) with one key per rule as described in the initial post by @reinholdsson could be useful.

It's definitely already possible to do this, but could be worth considering making a more user-friendly way of doing this if the demand is high enough.

borchero:
@danielcrane these days, dataframely provides `FilterResult.details()`. Does that do what you need?

ozanozbeker:
I just came across this thread. I ended up in the same place from my own work: I wanted Dataframely rules to show up as Dagster asset checks, hit exactly what @reinholdsson describes here, and have been building it out publicly as [dagster-dataframely](https://github.com/ozanozbeker/dagster-dataframely). The README serves as the user guide right now but I'm working on a docs site, the following are pulled from it:

One decorator attaches a schema to a Dagster asset:

```python
import dagster as dg
[...]

Issue #190, opened 2025-10-23
Feature Request: Add method(s) to `FailureInfo` to get a dataframe, which show more explicitly what rule(s) each row broke


DeflateAwning:
The goal is to be able to write out a dataframe to a file with more details of what's broken.

I'll propose a few options. Most of them make sense to implement simultaneously imo, though feel free to pick your favourite.

### Option 1: `failure_info.invalid(keep_rule_columns: bool = False)`

Simply make the `.drop(self._rule_columns)` part conditional. Not too sure what the names of those columns are, but they're probably reasonable.

### Option 2: `failure_info.invalid(with_rule_list: Literal['json', 'csv', 'list'] | None = None)`

Add a column  to the dataframe called something like `dataframely_failed_rules`. It would then add the column as a json list, as a comma-separated list of string, or as a `pl.List(pl.String)`, depending on the option.

delsner:
> The goal is to be able to write out a dataframe to a file with more details of what's broken.

Could you please elaborate why a regular `FailureInfo.write_parquet` (properly supported since #75) doesn't suffice your needs?

I'd be hesitant to convert error details, to e.g. CSV format within dataframely. I think this is something the user should be responsible for.

Issue #124, opened 2025-08-13
Get rules that failed in each row


elbg:
After seeing the output of failure.counts in 

```python
good, failure = MySchema.filter(df, cast=True)
failure.counts() # {'a': 128, 'n': 540}
```
I would like to be able to filter the rows that break each rule. I did not find that in the docs, is it already available ? 

Thanks !
Matthieu

borchero:
I think this was solved in #275.

PR #275, opened 2026-02-16
feat: Expose a user-friendly version of `FailureInfo._lf` in `FailureInfo.details()`

AndreasAlbertQC:
# Motivation

Addresses #272, #122, #124, #190

# Changes

* Added `FailureInfo.violation_details`, which provides the same rows as `FailureInfo.invalid` but also includes rule columns as well as original row values. To make usage easier for users, I decided to cast True/False/None to an enum "valid"/"invalid"/"unknown" (Since I know from personal experience that I often forgot whether "True" means good or bad here)
* Updated docs and fixed pre-existing `rooom` typo

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,11 +1,13 @@
 version: 2
 build:
-  os: ubuntu-20.04
+  os: ubuntu-24.04
   tools:
     python: mambaforge-latest
   commands:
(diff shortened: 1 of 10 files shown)


AndreasAlbertQC:
docs build appears to be broken independently of this PR, see #276

borchero (changes requested):

borchero (approved):
Thanks!

PR #283, opened 2026-02-20
docs: Update resolution docstrings to reference truncate instead of round

haydena7:
# Summary

In `dataframely/columns/datetime.py`, the docstrings for the `Date`, `Time`, `Datetime`, and `Duration` classes state that the resolution argument uses the formatting language of the `polars` datetime `round` method.

This is semantically confusing, as the actual underlying implementation uses the `polars` datetime `truncate` method.

# Motivation

While both methods share the same formatting strings, their behavior is materially different. The `round` method maps values in the second half of an interval to the end of their bucket, whereas `truncate` maps each value to the start of its bucket. Updating the docstring to reference truncate accurately reflects the strict boundary validation happening under the hood.

# Changes

Updated docstrings in `Date`, `Time`, `Datetime`, and `Duration` to replace "`round` method" with "`truncate` method".

### DIFF
--- a/dataframely/columns/datetime.py
+++ b/dataframely/columns/datetime.py
@@ -60,7 +60,7 @@ class Date(OrdinalMixin[dt.date], Column):
             max_exclusive: Like `max` but exclusive. May not be specified if `max`
                 is specified and vice versa.
             resolution: The resolution that dates in the column must have. This uses the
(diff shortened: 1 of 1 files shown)


borchero (approved):
Thanks!

PR #290, opened 2026-03-02
docs: Add coding agent docs, `SKILL.md`, and update copilot instructions

AndreasAlbertQC:
# Motivation

We want to provide instructions for using `dataframely` with AI coding agents.

# Changes


SKILL.md
```diff
@@ -0,0 +1,130 @@
...
+
+A `dy.Collection` describes a set of related dataframes, each described by a `dy.Schema`. Dataframes in a collection
+should share at least a subset of their primary key.
+
+```python
+class MyStreetSchema(dy.Schema):
+    """A schema for a dataframe describing streets."""
+
+    # Shared primary key component with MyHouseSchema
+    street: dy.String(primary_key=True)
```
delsner:
```suggestion
    street = dy.String(primary_key=True)
```
This should be an assignment, same below.

delsner (approved):
Thanks, I think this is a good starting point. Just a few nits.

borchero (approved):

PR #291, opened 2026-03-03
fix: Raise on duplicated aliases

gab23r:
Raises an `ImplementationError` when a schema contains columns with duplicate aliases.

# Motivation

I was using LLM to generate dataframely schemas, it gives me duplicated alias like in:

```python
class MySchema(dy.Schema):
    a = dy.Int64(alias="a")
    b = dy.Int64(alias="a")
```
It could be nice from dataframely it icould raise in this case.

# Changes

- Detect duplicate aliases within a single schema
- Detect duplicate aliases when inheriting from a parent schema
- Add tests for both cases

### DIFF
--- a/dataframely/_base_schema.py
+++ b/dataframely/_base_schema.py
@@ -84,6 +84,10 @@ class Metadata:
     rules: dict[str, RuleFactory] = field(default_factory=dict)
 
     def update(self, other: Self) -> None:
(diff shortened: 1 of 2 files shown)


AndreasAlbertQC (approved):
Thanks @gab23r !

PR #301, opened 2026-03-23
feat: Allow specifying Duration time unit

nardi:
# Motivation

I tried to use the library, but our data model requires `Duration`s with a specific time unit. I saw this was implemented for `Datetime` but not for `Duration`, so I implemented it in an analogous way.

# Changes

1. Added `time_unit` attribute to `dataframely.columns.datetime.Duration`.
2. Use `time_unit` when converting to Polars and Arrow data types.
3. Use `time_unit` when generating samples.
4. Add test for conversions to `pyarrow` schema, analogous to `Datetime`.


borchero (approved):
Lgtm, thanks!

PR #322, opened 2026-04-09
feat(columns): Add methods to transform existing column definitions

runkelcorey:
# Motivation

Suppose you have a schema like:
```python
class MySchema(dy.Schema):
    col1 = dy.Struct(inner = {"foo": dy.List(inner = dy.Struct(inner = {"bar": dy.String}))}, alias = "my.reserved.column.alias")
```
and now I want to add a new column that's basically identical, just with a different alias. That's a lot to type again! Plus, if the definitions change together, it's easier for them to grow out of sync. Instead, of redefining this column, I'd like to do this:
```python
class MySchema(dy.Schema):
    col1 = dy.Struct(inner = {"foo": dy.List(inner = dy.Struct(inner = {"bar": dy.String}))}, alias = "my.reserved.column.alias")
    col2 = col1.with_alias("my.other.reserved.alias")
```
Same goes for other common properties like `primary_key`, `nullability`, and `metadata`. At @mbta, we use dataframely as part of our ingestion ETL. We ingest deeply nested JSON and being able to reuse column definitions would be helpful.

# Changes

1. Add a new method `with_properties` that returns a new column with defined properties. For example, `dy.List(inner = dy.Any()).with_properties(inner = dy.String())` would return a list with a `String` inner column. All public properties are settable, though invalid properties (eg. `Any` with `nullable = False` are still disallowed since this implementation works through the constructor.
2. Add wrappers for common properties `with_nullable`, `with_alias`, `with_metadata`, and `with_primary_key`
3. Enable `Any` to accept `nullable` and `primary_key` kwargs. This is consistent with how `Array` works and allows a much cleaner implementation of `with_properties`

### DIFF
--- a/dataframely/columns/_base.py
+++ b/dataframely/columns/_base.py
@@ -235,20 +235,18 @@ class Column(ABC):
         return pl.col(self.name)
 
     def with_properties(self, **kwargs: Any) -> Self:
-        """Return a new column definition with updated public properties.
+        """Copy the current column definition while updating the provided properties.
 
-        Creates a copy of this column with the specified properties updated.
         All other properties from the original column are preserved.
 
         Args:
-            **kwargs: Properties to update on the new column instance.
-
+            **kwargs: Properties to update on the new column instance. The set of allowed properties depends on the type of the column.
         Returns:
             A new column instance with updated properties.
         """
         new_kwargs = {
-            k: v for k, v in self.__dict__.items() if not k.startswith("_")
-        } | kwargs  # keep private attributes private
+            k: getattr(self, k) for k in inspect.signature(self.__class__).parameters
+        } | kwargs
         return self.__class__(**new_kwargs)
 
     def with_nullable(self, nullable: bool) -> Self:
--- a/dataframely/columns/_base.py
+++ b/dataframely/columns/_base.py
@@ -241,6 +241,7 @@ class Column(ABC):
 
         Args:
             **kwargs: Properties to update on the new column instance. The set of allowed properties depends on the type of the column.
+
         Returns:
             A new column instance with updated properties.
         """
--- a/dataframely/columns/_base.py
+++ b/dataframely/columns/_base.py
@@ -234,6 +234,78 @@ class Column(ABC):
         """Obtain a Polars column expression for the column."""
         return pl.col(self.name)
 
+    def with_properties(self, **kwargs: Any) -> Self:
+        """Return a new column definition with updated public properties.
+
+        Creates a copy of this column with the specified properties updated.
+        All other properties from the original column are preserved.
+
+        Args:
+            **kwargs: Properties to update on the new column instance.
+
+        Returns:
+            A new column instance with updated properties.
+        """
+        new_kwargs = {
+            k: v for k, v in self.__dict__.items() if not k.startswith("_")
+        } | kwargs  # keep private attributes private
+        return self.__class__(**new_kwargs)
+
+    def with_nullable(self, nullable: bool) -> Self:
+        """Return a new column definition with specified nullability.
+
+        Args:
+            nullable: Whether the new column may contain null values.
+
+        Returns:
+            A new column instance with updated nullability.
+        """
+        return self.with_properties(nullable=nullable)
+
+    def with_alias(self, alias: str) -> Self:
+        """Return a new column definition with a specified alias.
+
+        Args:
+            alias: The alias to use for the column name.
+
+        Returns:
+            A new column instance with the specified alias.
+        """
+        return self.with_properties(alias=alias)
+
+    def with_check(self, check: Check) -> Self:
+        """Return a new column definition with a specified check.
+
+        Args:
+            check: A custom validation rule or rules for the column.
+
+        Returns:
+            A new column instance with the specified check.
+        """
+        return self.with_properties(check=check)
+
+    def with_primary_key(self, primary_key: bool) -> Self:
+        """Return a new column definition with a specified primary key status.
+
+        Args:
+            primary_key: Whether the column should be part of the primary key.
+
+        Returns:
+            A new column instance with updated primary key status.
+        """
+        return self.with_properties(primary_key=primary_key)
+
+    def with_metadata(self, metadata: dict[str, Any]) -> Self:
+        """Return a new column definition with specified metadata.
+
+        Args:
+            metadata: A dictionary of metadata to attach to the column.
+
+        Returns:
+            A new column instance with the specified metadata.
+        """
+        return self.with_properties(metadata=metadata)
+
     # ----------------------------------- SAMPLING ----------------------------------- #
 
     def sample(self, generator: Generator, n: int = 1) -> pl.Series:
--- a/dataframely/columns/any.py
+++ b/dataframely/columns/any.py
@@ -25,18 +25,12 @@ class Any(Column):
     def __init__(
         self,
         *,
-        nullable: bool = True,
-        primary_key: bool = False,
         check: Check | None = None,
         alias: str | None = None,
         metadata: dict[str, Any] | None = None,
     ):
         """
         Args:
-            nullable: Whether this column may contain null values.
-                Always `True` for the Any type.
-            primary_key: Whether this column is part of the primary key of the schema.
-                Always `False` for the Any type.
             check: A custom rule or multiple rules to run for this column. This can be:
                 - A single callable that returns a non-aggregated boolean expression.
                 The name of the rule is derived from the callable name, or defaults to
@@ -54,10 +48,6 @@ class Any(Column):
                 names, the specified alias is the only valid name.
             metadata: A dictionary of metadata to attach to the column.
         """
-        if not nullable:
-            raise NotImplementedError("The 'Any' type column must be nullable.")
-        if primary_key:
-            raise NotImplementedError("The 'Any' type column cannot be a primary key.")
         super().__init__(
             nullable=True,
             primary_key=False,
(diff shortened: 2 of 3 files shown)


dataframely/columns/any.py
```diff
@@ -25,12 +25,18 @@ class Any(Column):
     def __init__(
         self,
         *,
+        nullable: bool = True,
+        primary_key: bool = False,
```
AndreasAlbertQC: Why do we need to modify the signature for `Any`?

tests/columns/test_base.py
```diff
@@ -1,12 +1,184 @@
...
+        "UInt32",
+        "UInt64",
+    ],
+)
+def test_with_properties(existing_col: dy.Column, properties: dict[str, Any]) -> None:
+    """It assigns the property to the new column but not the old column."""
+    new_col = existing_col.with_properties(**properties)
+
+    assert all(getattr(new_col, key) == value for key, value in properties.items())
+    assert all(getattr(existing_col, key) != value for key, value in properties.items())
```
AndreasAlbertQC:
```suggestion
    assert all(getattr(existing_col, key) != value for key, value in properties.items()), "Sanity check for test setup failed: Properties to be changed are not different from original state."
    assert all(getattr(new_col, key) == value for key, value in properties.items()), "Property was not updated correctly"
    
```

tests/columns/test_base.py
```diff
@@ -1,12 +1,184 @@
...
+    assert all(
+        getattr(existing_col, key) == getattr(new_col, key)
+        for key in existing_col.__dict__.keys()
+        if key not in properties
+    )
```
AndreasAlbertQC:
```suggestion
    assert all(
        getattr(existing_col, key) == getattr(new_col, key)
        for key in existing_col.__dict__.keys()
        if key not in properties
    ), "Property was updated even though it should not have been"
```

AndreasAlbertQC: Let's add an assertion to verify that the original column object was not changed

tests/columns/test_base.py
```diff
@@ -1,12 +1,184 @@
...
+    ],
+)
+def test_with(
+    col_type: type[dy.Column],
+    col_kwargs: dict[str, Any],
+    property: str,
+    original_value: str | dict[str, Any] | Check | bool,
+    new_value: str,
+) -> None:
+    """It only updates the called property."""
```
AndreasAlbertQC:
```suggestion
```

see above and also: the test verifies other aspects, too, right?

tests/columns/test_base.py
```diff
@@ -1,12 +1,184 @@
...
+    # Some column types don't support primary_key
+    if property == "primary_key" and col_type in [dy.Any, dy.Array, dy.List, dy.Object]:
+        pytest.xfail(f"{col_type.__name__} does not support primary_key")
+
+    # Any column type doesn't support changing nullable
+    if property == "nullable" and col_type == dy.Any:
+        pytest.xfail("Any does not support changing nullable")
+
+    col = col_type(**col_kwargs)
+    setattr(col, property, original_value)
```
AndreasAlbertQC: Instead of calling `setattr` here, should `property` and `original_value` be added to `col_kwargs` and passed on the line above?

AndreasAlbertQC (changes requested):
Thanks for the PR @runkelcorey! I think this quite useful and the code already looks great. I mostly have suggestions on testing

AndreasAlbertQC:
Sorry wrong button

dataframely/columns/any.py
runkelcorey: otherwise, we can't use the constructor to pass the kwargs. The constructor is cleaner and the alternative—setting properties directly—would allow a user to set `nullable = False` on an `Any` type

dataframely/columns/any.py
```diff
@@ -25,12 +25,18 @@ class Any(Column):
...
         check: Check | None = None,
         alias: str | None = None,
         metadata: dict[str, Any] | None = None,
     ):
```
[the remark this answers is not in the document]
runkelcorey: I like the idea of raising an error (perhaps `NotImplementedError`?) when a user passes an incompatible value for `nullable` or `primary_key`.

tests/columns/test_base.py
runkelcorey: good call

tests/columns/test_base.py
```diff
@@ -1,12 +1,184 @@
...
+def test_with(
+    col_type: type[dy.Column],
+    col_kwargs: dict[str, Any],
+    property: str,
+    original_value: str | dict[str, Any] | Check | bool,
+    new_value: str,
+) -> None:
+    """It only updates the called property."""
+    # Some column types don't support primary_key
+    if property == "primary_key" and col_type in [dy.Any, dy.Array, dy.List, dy.Object]:
```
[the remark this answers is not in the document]
runkelcorey: would we rather these all raise errors? that may be more helpful for users

dataframely/columns/_base.py
```diff
@@ -234,6 +234,78 @@ def col(self) -> pl.Expr:
...
+
+        Args:
+            **kwargs: Properties to update on the new column instance.
+
+        Returns:
+            A new column instance with updated properties.
+        """
+        new_kwargs = {
+            k: v for k, v in self.__dict__.items() if not k.startswith("_")
+        } | kwargs  # keep private attributes private
```
AndreasAlbertQC:
```suggestion
        valid_parameters = set(inspect.signature(self.__class__).parameters.keys())
        new_kwargs = {
            k: v for k, v in self.__dict__.items() if k in valid_parameters
        } | kwargs
```

What do you think about this? That would allow us to keep nullable and primary_key out of the `Any` constructor signature, which I think is neat

dataframely/columns/_base.py
```diff
@@ -234,6 +234,78 @@ def col(self) -> pl.Expr:
         """Obtain a Polars column expression for the column."""
         return pl.col(self.name)
 
+    def with_properties(self, **kwargs: Any) -> Self:
+        """Return a new column definition with updated public properties.
+
+        Creates a copy of this column with the specified properties updated.
+        All other properties from the original column are preserved.
```
borchero:
```suggestion
        """Copy the current column definition while updating the provided properties.

        All other properties from the original column are preserved.
```

dataframely/columns/_base.py
```diff
@@ -234,6 +234,78 @@ def col(self) -> pl.Expr:
...
+
+        Args:
+            **kwargs: Properties to update on the new column instance.
```
borchero:
```suggestion
            **kwargs: Properties to update on the new column instance. The set of allowed properties depend on the type of the column.
```

dataframely/columns/_base.py
borchero: You don't even need to iterate over `__dict__`:

```suggestion
        new_kwargs = {
            k: getattr(self, k) for k in inspect.signature(self.__class__).parameters
        } | kwargs
```

dataframely/columns/_base.py
```diff
@@ -234,6 +234,78 @@ def col(self) -> pl.Expr:
...
+
+        Returns:
+            A new column instance with updated properties.
+        """
+        new_kwargs = {
+            k: v for k, v in self.__dict__.items() if not k.startswith("_")
+        } | kwargs  # keep private attributes private
+        return self.__class__(**new_kwargs)
+
+    def with_nullable(self, nullable: bool) -> Self:
```
borchero, line 253: Are all of these special methods really necessary if we have `with_properties`?

dataframely/columns/any.py
borchero: With the change above where we don't iterate over `__dict__`, this won't be necessary :)

borchero (commented):
Thanks for the initiative @runkelcorey!

dataframely/columns/_base.py
runkelcorey: Necessary? No. I'd argue they're ergonomic tho. I'm happy if y'all wanna be conservative and just admit `with_properties`

borchero: > I'd argue they're ergonomic

Yeah fair, I think I'm mainly worrying about "consistency". I.e. why's there no `with_min`, `with_regex`, ...

Wdyt @AndreasAlbertQC?

runkelcorey: If I can offer an easy answer for consistency, it's that these are the kwargs defined in `Column`

AndreasAlbertQC: I also thought about this, but then thought that it's really free syntactic sugar. Can't really see a practical downside

AndreasAlbertQC (approved):
Thanks @runkelcorey ! LGTM

dataframely/columns/_base.py
borchero: Ok, fine by me :)

dataframely/columns/_base.py
```diff
@@ -234,6 +234,76 @@ def col(self) -> pl.Expr:
...
 
+    def with_properties(self, **kwargs: Any) -> Self:
+        """Copy the current column definition while updating the provided properties.
+
+        All other properties from the original column are preserved.
+
+        Args:
+            **kwargs: Properties to update on the new column instance. The set of allowed properties depends on the type of the column.
+        Returns:
+            A new column instance with updated properties.
```
borchero, line 246:
```suggestion
        Args:
            **kwargs: Properties to update on the new column instance. The set of allowed properties depends on the type of the column.

        Returns:
            A new column instance with updated properties.
```

borchero (approved):
Thanks!

PR #325, opened 2026-04-13
feat: Add `unique` rule to `dy.Column`

### LINKED ISSUE #313: Feature Request: `unique=True` column constraint
Currently, enforcing uniqueness on a column requires either:

1. `primary_key=True` - but when multiple columns have this, they form a **composite key** (uniqueness checked on the combination, not individually)
2. A custom `@dy.check` function - works, but is verbose for such a common constraint

There's no convenient inline way to express "this column must contain unique values" independently.

### Example

```python
import dataframely as dy
import polars as pl

class Users(dy.Schema):
    id = dy.Int64(primary_key=True)
    email = dy.String(primary_key=True)  # Want unique emails, but this creates composite key (id, email)
    username = dy.String()

# This passes validation, but shouldn't:
df = pl.DataFrame({
    "id": [1, 2],
    "email": ["a@x.com", "a@x.com"],  # Duplicate email!
    "username": ["alice", "alex"], 
})
Users.validate(df)  # No error - composite key (1, "a@x.com") != (2, "a@x.com")
```

### Proposed Solution

Add a `unique=True` parameter to column definitions:

```python
class Users(dy.Schema):
    id = dy.Int64(primary_key=True)
    email = dy.String(unique=True)     # Must be unique independently
    username = dy.String(unique=True)  # Must be unique independently
```

This would check each column's uniqueness separately from the primary key constraint, without requiring a custom check function.

gab23r:
# Motivation

Closes #313 

# Changes

Add the new rule using the same logic than primary_keys.

Drive by:
- Allow `primary_keys` for array dtype as it now works in polars, I added a test of it.
- Disallow  `primary_keys` for object dtype as it never worked. (technically breaking but IMHO shouldn't break anything)

### DIFF
--- a/dataframely/_base_schema.py
+++ b/dataframely/_base_schema.py
@@ -39,7 +39,14 @@ def _build_rules(
     # Add primary key validation to the list of rules if applicable
     primary_key = _primary_key(columns)
     if len(primary_key) > 0:
-        rules["primary_key"] = Rule(~pl.struct(primary_key).is_duplicated())
+        rules["primary_key"] = Rule(pl.struct(primary_key).is_unique())
+
+    # Add unique column validation rules
+    unique_columns = _unique_columns(columns)
+    for col_name in unique_columns:
+        # wrap the column in a struct to make `is_unique` work with list/arrays
+        # https://github.com/pola-rs/polars/issues/27286
+        rules[f"{col_name}|unique"] = Rule(pl.struct(col_name).is_unique())
 
     # Add column-specific rules
     column_rules = {
@@ -69,6 +76,10 @@ def _build_rules(
 
 def _primary_key(columns: dict[str, Column]) -> list[str]:
     return list(k for k, col in columns.items() if col.primary_key)
+
+
+def _unique_columns(columns: dict[str, Column]) -> list[str]:
+    return list(k for k, col in columns.items() if col.unique)
 
 
 # ------------------------------------------------------------------------------------ #
@@ -299,6 +310,11 @@ class BaseSchema(metaclass=SchemaMeta):
     def primary_key(cls) -> list[str]:
         """The primary key columns in this schema (possibly empty)."""
         return _primary_key(cls.columns())
+
+    @classmethod
+    def unique_columns(cls) -> list[str]:
+        """The columns with unique constraints in this schema (possibly empty)."""
+        return _unique_columns(cls.columns())
 
     @classmethod
     def _validation_rules(cls, *, with_cast: bool) -> dict[str, Rule]:
--- a/dataframely/columns/object.py
+++ b/dataframely/columns/object.py
@@ -22,7 +22,6 @@ class Object(Column):
         self,
         *,
         nullable: bool = False,
-        primary_key: bool = False,
         check: Check | None = None,
         alias: str | None = None,
         metadata: dict[str, Any] | None = None,
@@ -30,7 +29,6 @@ class Object(Column):
         """
         Args:
             nullable: Whether this column may contain null values.
-            primary_key: Whether this column is part of the primary key of the schema.
             check: A custom rule or multiple rules to run for this column. This can be:
                 - A single callable that returns a non-aggregated boolean expression.
                 The name of the rule is derived from the callable name, or defaults to
@@ -50,7 +48,6 @@ class Object(Column):
         """
         super().__init__(
             nullable=nullable,
-            primary_key=primary_key,
             check=check,
             alias=alias,
             metadata=metadata,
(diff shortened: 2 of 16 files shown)


dataframely/columns/object.py
```diff
@@ -22,15 +22,13 @@ def __init__(
         self,
         *,
         nullable: bool = True,
-        primary_key: bool = False,
         check: Check | None = None,
         alias: str | None = None,
```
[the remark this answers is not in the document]
AndreasAlbertQC: Was there a reason to change this?

tests/schema/test_sample.py
```diff
@@ -261,3 +261,35 @@ def test_sample_override_sequence_with_missing_keys_and_resampling() -> None:
...
+    email = dy.String(unique=True)
+
+
+@pytest.mark.parametrize("n", [0, 100])
+def test_sample_unique_constraint(n: int) -> None:
+    df = UniqueSchema.sample(n, generator=Generator(seed=42))
+    assert len(df) == n
+    UniqueSchema.validate(df)
+    # Verify uniqueness
+    assert df.get_column("email").n_unique() == n
```
AndreasAlbertQC: I think this test would be more meaningful if we used a smaller dtype here, e.g. `UInt8`. With a string, I think your chance of sampling two identical strings is very small either way, it's unclear if the unique constraint worked.

tests/schema/test_sample.py
```diff
@@ -261,3 +261,35 @@ def test_sample_override_sequence_with_missing_keys_and_resampling() -> None:
...
+
+
+class MultiUniqueSchema(dy.Schema):
+    a = dy.Int64(unique=True)
```
AndreasAlbertQC: same here

AndreasAlbertQC (changes requested):
Thanks @gab23r ! This looks quite nice to me already. Small suggestions.

dataframely/columns/object.py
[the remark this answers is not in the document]
gab23r: this is more a pass-by changes, I though about NOT adding is_unique for object as it is not allowed by polars and then I saw the primary_key argument here... It shouldn't be here, as primary_key for object dtype are not possible.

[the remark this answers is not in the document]
AndreasAlbertQC: You're right, I was not aware of the polars behavior here. TLDR:
1. You can call `df.unique()` on a dataframe with an object column and it will work as expected iff your objects are hashable
2. You cannot call `df.select(pl.col("mycol").is_unique())`, which errors saying that uniqueness checks are not supported for object columns.

In any case, passing `primary_key=True` to `dy.Object` leads to failing validation no matter what, so I think it's reasonable to remove it here.

tests/schema/test_sample.py
gab23r: I used Int16, with Int8 the test was not even feasible 😄

AndreasAlbertQC: Ok for future reference I had to pull up a birthday paradox calculator to convince myself that I think this is OK :D

AndreasAlbertQC (approved):
Thanks @gab23r !

dataframely/_base_schema.py
```diff
@@ -39,7 +39,14 @@ def _build_rules(
     # Add primary key validation to the list of rules if applicable
     primary_key = _primary_key(columns)
     if len(primary_key) > 0:
-        rules["primary_key"] = Rule(~pl.struct(primary_key).is_duplicated())
+        rules["primary_key"] = Rule(pl.struct(primary_key).is_unique())
```
borchero, line 42: Meta-comment because I just noticed: `pl.struct(primary_key).is_unique()` is likely much more inefficient than `pl.col(primary_key).is_unique()` if we only have a single primary key. We might want to introduce an optimization for this after benchmarking :smile:

dataframely/_base_schema.py
```diff
@@ -39,7 +39,14 @@ def _build_rules(
...
+
+    # Add unique column validation rules
+    unique_columns = _unique_columns(columns)
+    for col_name in unique_columns:
+        # wrap the column in a struct to make `is_unique` work with list/arrays
+        # https://github.com/pola-rs/polars/issues/27286
+        rules[f"{col_name}|unique"] = Rule(pl.struct(col_name).is_unique())
```
borchero, line 49: Sorry I didn't review previously but I find this implementation suboptimal. Why is it on the schema if we do not check composite uniqueness but uniqueness of individual columns? This should be on the column which would also allow for much more efficient evaluation of `is_unique` for primitive types (because we can very easily skip the struct-wrapping).

borchero: Besides, this also breaks for nested types; for example setting `unique` on a list element is simply ignored.

PR #326, opened 2026-04-13
fix: Default to non nullable column for array and object

gab23r:
I think is as been missed when merging the [PR](https://github.com/Quantco/dataframely/pull/174).
Or maybe it wasn't possible to implement at the time. In any case I don't see why object and array are nullable by default.

This is breaking, how do you want to proceed ?


borchero (approved):
Thanks for the fix, this was surely an oversight! Can you still fix the tests?

I think we'll have to mark that as non-breaking as I don't want this to cause v3. It's obviously a bug though.

PR #330, opened 2026-04-17
fix: Propagate column overrides through grandchild schemas

### LINKED ISSUE #329: Schema column override breaks in grandchild class (regression in 2.8.0)
Repro:
```python
import dataframely as dy

class Base(dy.Schema):
    amt = dy.Float64(nullable=True)

class Child(Base):
    amt = dy.Float64(nullable=False)

class Grandchild(Child):
    pass
# ImplementationError: Columns {'amt'} are duplicated with conflicting definitions.
```

Expected: Grandchild inherits Child's override of amt without error, the same way two-level inheritance (Base → Child) works.

Actual: `_get_metadata_recursively` walks the MRO and sees amt on both Child and Base with different definitions, raising `ImplementationError`. It doesn't account for the fact that Child's definition should shadow Base's.

Bisect: Works in >=2.7.0,<2.8.0. Broken in >=2.8.0,<2.9.0 and 2.9.1.

AndreasAlbertQC:
# Motivation

Fixes #329. Starting in 2.8.0, defining a grandchild schema that inherits from a child which overrides a parent column raises `ImplementationError: Columns {...} are duplicated with conflicting definitions.`:

```python
class Base(dy.Schema):
    amt = dy.Float64(nullable=True)

class Child(Base):
    amt = dy.Float64(nullable=False)

class Grandchild(Child):  # ImplementationError
    pass
```

`SchemaMeta.__new__` drops parent columns shadowed by the current namespace via `_remove_overridden_columns`, but `_get_metadata_recursively` — used when walking bases during grandchild creation — never applied the same logic. So both `Child`'s override and `Base`'s original `amt` landed in the merged metadata and tripped the conflict check.

# Changes

- Call `_remove_overridden_columns` inside `_get_metadata_recursively` so column overrides propagate through inheritance chains of any depth.
- Add regression tests covering grandchild and great-grandchild inheritance of overridden columns.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dataframely/_base_schema.py
+++ b/dataframely/_base_schema.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import sys
 import textwrap
 from abc import ABCMeta
+from collections.abc import Mapping
 from copy import copy
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any
@@ -207,7 +208,7 @@ class SchemaMeta(ABCMeta):
     @staticmethod
     def _remove_overridden_columns(
         result: Metadata,
-        namespace: dict[str, Any],
+        namespace: Mapping[str, Any],
         bases: tuple[type[object], ...],
     ) -> None:
         """Remove inherited columns that the child namespace explicitly overrides.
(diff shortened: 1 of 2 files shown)


borchero (approved):
Thanks! Approval modulo coverage

PR #333, opened 2026-04-21
fix: Make `unique` work for nested types and improve performance

borchero:
# Motivation

Follow-up to #325, see https://github.com/Quantco/dataframely/pull/325#discussion_r3120678573.

# Changes

- Move `unique` to the column definitions as it is column- rather than schema-business
- Remove the `unique_columns` classmethod with the same argument
- Use `pl.col("...").is_unique()` by default instead of wrapping in a struct for improved performance
- Add tests for nested uniqueness and uniqueness of complex types

### DIFF
--- a/dataframely/_base_schema.py
+++ b/dataframely/_base_schema.py
@@ -42,13 +42,6 @@ def _build_rules(
     if len(primary_key) > 0:
         rules["primary_key"] = Rule(pl.struct(primary_key).is_unique())
 
-    # Add unique column validation rules
-    unique_columns = _unique_columns(columns)
-    for col_name in unique_columns:
-        # wrap the column in a struct to make `is_unique` work with list/arrays
-        # https://github.com/pola-rs/polars/issues/27286
-        rules[f"{col_name}|unique"] = Rule(pl.struct(col_name).is_unique())
-
     # Add column-specific rules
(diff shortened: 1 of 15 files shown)


AndreasAlbertQC (approved):

gab23r:
Indeed much cleaner than my implementation ! 
Just for fun.  Wapping `in_unique` in `pl.struct` is actually faster on micro benchmark  😄 

```python
import dataframely as dy
import polars as pl


class Schema(dy.Schema):
    a = dy.Int64()
df = Schema.sample(1000000)

%timeit _ = df.select(pl.struct("a").is_unique())
# 8.08 ms ± 495 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit _ = df.select(pl.col("a").is_unique())
# 28.6 ms ± 1.04 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
This is because `pl.struct("a").is_unique()` dispatch to `df.select("a").unnest().is_unique()` which is exectuted in parallel. Which is not the case for `Expr.is_unique`.

borchero:
Thanks for the benchmark @gab23r, this is really surprising 👀 but good to know that this behavior changed since I last benchmarked this >1y ago 😅

PR #335, opened 2026-04-27
feat: Add `DataFrame` support in `dy.Collection`

### LINKED ISSUE #319: Feature Request: `dy.DataFrame` Support in Collections
Allow Collections to declare `dy.DataFrame` members alongside `dy.LazyFrame`:

```python
class MyCollection(dy.Collection):
    users: dy.DataFrame[UserSchema]      # Eager
    orders: dy.DataFrame[OrderSchema]    # Eager
```

## Motivation

- **Small datasets**: When working with small data, `.collect()` is unnecessary ceremony
- **EDA workflows**: Inspecting results in notebooks/REPLs is more convenient with DataFrames
- **Consistency**: Schema validation already supports both `df` and `lf` - Collections should too

```python
# Current (verbose for small data)
collection = MyCollection.validate(data)
collection.users.collect()  # extra step just to see data

# Proposed (direct)
collection = MyCollection.validate(data)
collection.users  # already a DataFrame

### PR DESCRIPTION
**Pull request by gab23r**
# Motivation

Closes #319 

# Changes

- Add `is_lazy` attribute in `MemberInfo`.
- Add `lazy_members` and `eager_members` to `dy.Collection`- 
- Internals still works on lazyframes via the new `_to_lazy_dict` methods.
- Modify `Collection._init` to collect the dataframe is the eager case.

### CODE REVIEW

File: dataframely/collection/_base.py [hunk 1]
```diff
@@ -241,39 +244,46 @@ def _derive_member_info(
...
-            not_none_args = [arg for arg in union_args if get_origin(arg) is not None]
-            if len(not_none_args) == 0 or not issubclass(
-                get_origin(not_none_args[0]), TypedLazyFrame
-            ):
+            # Get the non-None type (exactly one exists given prior checks)
+            not_none_arg = next(arg for arg in union_args if arg is not type(None))
+
+            frame_origin = get_origin(not_none_arg)
+            if frame_origin is None:
                 raise AnnotationImplementationError(attr, type_annotation)
```
borchero, line 260: Why did these need to change? It feels like it's the same thing rewritten?

dataframely/collection/_base.py
```diff
@@ -371,20 +391,40 @@ def common_primary_key(cls) -> list[str]:
     def _filters(cls) -> dict[str, Filter[Self]]:
         return getattr(cls, _FILTER_ATTR)
 
-    def to_dict(self) -> dict[str, pl.LazyFrame]:
-        """Return a dictionary representation of this collection."""
+    def to_dict(self) -> dict[str, FrameType]:
```
borchero: This signature forces the user to always specify the return type. I would keep `pl.LazyFrame` here and non-conditionally call `.lazy()` on the attribute. Wdyt?

dataframely/collection/_base.py
```diff
@@ -371,20 +391,40 @@ def common_primary_key(cls) -> list[str]:
...
                 setattr(out, member_name, None)
-            else:
+            elif member.is_lazy:
                 setattr(out, member_name, data[member_name].lazy())
+            else:
+                frame = data[member_name]
+                if isinstance(frame, pl.LazyFrame):
+                    setattr(out, member_name, frame.collect())
+                else:
+                    setattr(out, member_name, frame)
```
borchero: You can just do `.lazy().collect()` which works on both lazy and eager frames

dataframely/collection/collection.py
```diff
@@ -735,7 +735,7 @@ def join(
                     how=how,
                     maintain_order=maintain_order,
                 )
-                for key, lf in self.to_dict().items()
+                for key, lf in self._to_lazy_dict().items()
```
borchero: Might not be necessary with the comment above

dataframely/collection/collection.py
```diff
@@ -788,16 +788,15 @@ def collect_all(self) -> Self:
...
+            with :class:`~dataframely.DataFrame` are returned as DataFrames, while
+            members annotated with :class:`~dataframely.LazyFrame` are returned as
+            "shallow-lazy" frames (obtained by calling ``.collect().lazy()``).
         """
-        dfs = pl.collect_all(self.to_dict().values())
+        lazy_dict = self._to_lazy_dict()
+        dfs = pl.collect_all(lazy_dict.values())
         return self._init(
-            {key: dfs[i].lazy() for i, key in enumerate(self.to_dict().keys())}
+            {key: dfs[i].lazy() for i, key in enumerate(lazy_dict.keys())}
```
borchero: Is the `.lazy()` required?

borchero (commented):
Sorry it took so long to review this! Generally, I think that we can support this, I left a few comments to simplify the implementation a little

dataframely/collection/_base.py
gab23r: I change this because it was, IMO, missleading. 
The line `if not any(get_origin(arg) is None for arg in union_args)` is not actually checking that one arg is None, e.g. `get_origin(int) is None` So this example would pass this check:
```python
class MyCollection(dy.Collection):
    first: dy.DataFrame[MyTestSchema] | int
```

dataframely/collection/_base.py
gab23r: I agree that returning `FrameType` is not ideal. In the meantime, returning a Lazyframe where the user explicitly ask for a dataframe doesn't sound right as well IMO. I would prefer to return `pl.DataFrame | pl.LazyFrame`
User can himself use `.lazy`or `collect` if he wants specific type.

dataframely/collection/collection.py
gab23r: I don't follow, what is not necessary ?

gab23r: Ah ok, I understand after a small break 😁 
I think you are right, we can get rid of `_to_lazy_dict` and make `to_dict` return only lazyframe. It make internal typing much easier.

dataframely/collection/_base.py
```diff
@@ -372,9 +392,9 @@ def _filters(cls) -> dict[str, Filter[Self]]:
         return getattr(cls, _FILTER_ATTR)
 
     def to_dict(self) -> dict[str, pl.LazyFrame]:
```
borchero, line 394: Thanks for keeping this! Otherwise, this PR would also have resulted in a breaking change which, I think, would be rather undesirable for a change of this "magnitude".

borchero (approved):
Thanks for the updates @gab23r!

Issue #323, opened 2026-04-09
Feature request: Convert `dy.Schema` to `pydantic` model


kklein:
## Motivation

Various LLM providers allow for constrained decoding, making sure that their models generate responses adhering to a user-specified, structured output format -- see e.g. [Anthropic's API docs on structured outputs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs).

Not only is this possible on a REST API level. Client libraries have reflected this and packages such as [`openai`](https://developers.openai.com/api/docs/guides/structured-outputs#examples) or [`pydantic-ai`](https://pydantic.dev/docs/ai/api/pydantic-ai/agent/#pydantic_ai.agent.Agent.output_type) as well. These allow for specifying structured output formats via plain JSON, dicts or `pydantic` models.

If I want an LLM to generate data that I want to use as a DataFrame in the subsequent process steps, I would love to be able to tell the LLM it should generate data which adheres to a `dy.Schema` of mine, rather than having to translate (and potentially additionally maintain) a `pydantic` model which mirrors my `dy.Schema`, e.g. as having lists of row models.

## Target use case

1. I define a `dy.Schema` that is expected as input for a process stage of mine.
2. I call my LLM library to generate output that adheres to this schema by e.g. providing it the result of a `MySchema.to_pydantic` call as a generation argument.
3. I manually turn the `pydantic` object returned by my LLM client into a dataframe with a `MySchema.validate` call.

kklein:
Hi @borchero !

When using a `pydantic.BaseModel` to ask for structured output with `pydantic-ai`, descriptions of fields are given to the LLM for generation (either via prompt, tool or native constrained decoding), see [this article](https://pydantic.dev/docs/validation/latest/concepts/json_schema/#generating-json-schema). 

What would you say is the best way to try to get a description into the structured output specification when first converting from a `dy.Schema`?

[...]

PR #342, opened 2026-05-24
feat: Add `description` field to columns

### LINKED ISSUE #132: Support for docstring styled column documentation
Feature request to allow for columns level docstrings in the following form:
```python
class HouseSchema(dy.Schema):
    """Explanation of what my House Schema is really representing"""
    
    zip_code = dy.String(nullable=False, min_length=3)
    """A zip_code is a ..."""
    
    num_bedrooms = dy.UInt8(nullable=False)
    """Counts the rooms in a house not including the living room, kitchen and ..."""

    etc
```

This could potentially allow for printing of schema documentation notes into a standardised/centralised format (e.g., to support with creating data documentation)

borchero:
# Motivation

Implements https://github.com/Quantco/dataframely/issues/323#issuecomment-4505728567 and closes #132.

## Why introduce a `description` parameter?

Unfortunately, attribute docstrings cannot be extracted at runtime ([PEP 224](https://peps.python.org/pep-0224/) was rejected). Hence, this is the only way to make use of the description at runtime, e.g. to pass to `pydantic` field descriptions.

### DIFF
--- a/dataframely/columns/_base.py
+++ b/dataframely/columns/_base.py
@@ -453,7 +453,7 @@ class Column(ABC):
                     else getattr(self, param)
                 )
                 for param in inspect.signature(self.__class__.__init__).parameters
-                if param not in ("self", "alias")
+                if param not in ("self", "alias", "description")
             },
         }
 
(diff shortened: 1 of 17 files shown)


AndreasAlbertQC (changes requested):
Thanks @borchero ! Would it make sense to still extend tests by:
1. adding schemas with descriptions to the parametrization in `tests/schema/test_serialization.py`
2. adding schemas with description somewhere in `tests/schema/test_storage.py` ? 

The first would be to verify that serialization roundtrip still works, while the second also ensures that we do not produce new warnings when we should not

AndreasAlbertQC (approved):
Thanks!

danielcrane:
Have there been any discussions regarding also adding this functionality to rules?

borchero:
Hey @danielcrane, what would be the use case for that? The current descriptions are also descriptions for _columns_ rather than (column-level) _rules_.

PR #343, opened 2026-05-24
perf: Allow running as streaming node on the streaming engine

borchero:
# Motivation

I realized that, when running on the streaming engine, our current Rust plugin forces in-memory execution. In some simple benchmarks, this also improves execution speed on the streaming engine 10-20%. I did not measure memory consumption but would expect a significant drop.

**Example plan prior to this PR**

old

**Same plan after this PR**

new

### DIFF
--- a/dataframely/_plugin.py
+++ b/dataframely/_plugin.py
@@ -86,5 +86,10 @@ def all_rules_required(
         args=rules,
         kwargs={"null_is_valid": null_is_valid, "schema_name": schema_name},
         use_abs_path=True,
+        # NOTE: Conceptually, we're reducing the input to a single boolean value here.
+        #  However, we set this option to ensure that the plugin does not become
+        #  blocking on the streaming engine. A single boolean value is simply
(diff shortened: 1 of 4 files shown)


dataframely/_plugin.py
```diff
@@ -30,6 +30,7 @@ def all_rules_horizontal(rules: IntoExpr | Iterable[IntoExpr]) -> pl.Expr:
         function_name="all_rules_horizontal",
         args=rules,
         use_abs_path=True,
+        is_elementwise=True,
```
delsner, line 33: Can we somehow document that this is required for streaming mode to work?

delsner (approved):
LGTM. Is there a way to check in a test that the query plan does not fall back to in-memory execution mode if streaming mode is requested (e.g., checking if any node in the graph is of in-memory type)?

borchero:
> Is there a way to check in a test that the query plan does not fall back to in-memory execution mode if streaming mode is requested (e.g., checking if any node in the graph is of in-memory type)?

Good point, let me check

PR #353, opened 2026-06-02
fix: Forward storage_options to parquet metadata reads

### LINKED ISSUE #352: Parquet metadata read ignores storage_options, breaking typed reads from non-AWS S3 stores
### Summary
`Schema.read_parquet` / `read_from` (and `scan_parquet`, collection reads) fail
against S3-compatible stores reached via `storage_options` (lakeFS gateway, MinIO,
R2, Tigris, …). The **data** read gets `storage_options`, but the **embedded-schema
metadata** read does not — so polars falls back to `credential_provider="auto"` +
the default AWS S3 endpoint, hitting the wrong endpoint/creds.

### Repro
```python
opts = {  # lakeFS S3 gateway; any non-AWS S3 store reproduces this
    "aws_access_key_id": "…", "aws_secret_access_key": "…",
    "aws_endpoint_url": "https://lakefs.example.com",
    "aws_virtual_hosted_style_request": "false",
}
MySchema.read_parquet("s3://repo/branch/table.parquet", storage_options=opts)
# -> botocore TokenRetrievalError / AccessDenied: the call resolves the default
#    AWS chain + real AWS endpoint, not the store described by storage_options.
df = pl.read_parquet("s3://repo/branch/table.parquet", storage_options=opts)  # works
```

### Cause
`_storage/parquet.py`: `storage_options` arrives in `**kwargs` and is passed to
`pl.read_parquet` / `pl.scan_parquet`, but the metadata helpers call
`pl.read_parquet_metadata(path)` with no options:

```python
def read_frame(self, **kwargs):
    source = kwargs.pop("source")
    df = pl.read_parquet(source, **kwargs)        # storage_options
    metadata = _read_serialized_schema(source)    # dropped
```
`pl.read_parquet_metadata` accepts `storage_options=` — it's just not forwarded.

### Fix
Thread `kwargs.get("storage_options")` into the metadata helpers in
`dataframely/_storage/parquet.py`:

- `_read_serialized_schema` (L253–255) and `_read_serialized_collection` (L248–250):
  add `storage_options=None` param → `pl.read_parquet_metadata(path, storage_options=storage_options)`
- call sites: `read_frame` (L62), `scan_frame` (L56), `_collection_from_parquet`
  (L162/L167), `scan_failure_info` (L237)

(Threading `credential_provider` too would also fix auto-resolution.)

### Version
dataframely 2.10.1, polars <fill in>. Happy to open a PR.

mattijsdp:
# Motivation

Fixes https://github.com/Quantco/dataframely/issues/352.

`Schema.read_parquet`/`scan_parquet` (plus collection and failure-info reads) failed against S3-compatible stores reached via `storage_options`. The data read got `storage_options`, but the separate embedded-metadata read didn't — so polars fell back to the default AWS endpoint/credentials.

# Changes

- `_storage/parquet.py` — forward the storage options `pl.read_parquet_metadata` accepts (`storage_options`, `credential_provider`, `retries`) to every metadata read, via a small `_metadata_read_options(kwargs)` helper. Only present keys are forwarded, so defaults are unchanged.
- `read_parquet_metadata_schema` (`schema.py`) and `read_parquet_metadata_collection` (`collection/collection.py`) now accept and forward `**kwargs`.

Note: limited to polars parquet-metadata reads (covers the issue's single-file repro). A `storage_options`-only collection read still fails earlier, at fsspec-based member discovery (`url_to_fs`/`glob`), which uses a different option vocabulary than polars and would need a per-backend translation I think.

### DIFF
--- a/dataframely/_storage/parquet.py
+++ b/dataframely/_storage/parquet.py
@@ -252,16 +252,9 @@ class ParquetStorageBackend(StorageBackend):
 
 
 def _metadata_read_options(kwargs: dict[str, Any]) -> dict[str, Any]:
-    """Select the storage-related options to forward to a parquet metadata read.
-
-    The metadata is read separately from the data, via
-    :meth:`polars.read_parquet_metadata`. That function accepts only a narrow subset of
-    the options understood by :meth:`polars.read_parquet`/:meth:`polars.scan_parquet`,
-    so we cannot simply forward all ``kwargs``. We forward exactly the options that
-    determine *how the store is reached* so that metadata reads against non-AWS
-    S3-compatible stores (lakeFS, MinIO, R2, …) hit the correct endpoint and credentials
-    instead of falling back to the default AWS credential chain and endpoint.
-    """
+    # Forward only the options that `read_parquet_metadata` accepts (it has a narrower
+    # signature than `read_parquet`/`scan_parquet`) so the metadata read reaches the
+    # same store as the data read.
     return {
         key: kwargs[key]
         for key in ("storage_options", "credential_provider", "retries")
--- a/dataframely/_storage/parquet.py
+++ b/dataframely/_storage/parquet.py
@@ -147,6 +147,7 @@ class ParquetStorageBackend(StorageBackend):
         # between lazy and eager reads
         data = {}
         collection_types = []
+        metadata_options = _metadata_read_options(kwargs)
 
         fs: AbstractFileSystem = url_to_fs(path)[0]
         for key in members:
@@ -160,16 +161,14 @@ class ParquetStorageBackend(StorageBackend):
                 )
                 if is_file:
                     collection_types.append(
-                        _read_serialized_collection(
-                            source_path, **_metadata_read_options(kwargs)
-                        )
+                        _read_serialized_collection(source_path, **metadata_options)
                     )
                 else:
                     prefix = get_file_prefix(fs)
                     for file in fs.glob(fs.sep.join([source_path, "**", "*.parquet"])):
                         collection_types.append(
                             _read_serialized_collection(
-                                f"{prefix}{file}", **_metadata_read_options(kwargs)
+                                f"{prefix}{file}", **metadata_options
                             )
                         )
 
--- a/dataframely/_storage/parquet.py
+++ b/dataframely/_storage/parquet.py
@@ -53,13 +53,13 @@ class ParquetStorageBackend(StorageBackend):
     def scan_frame(self, **kwargs: Any) -> tuple[pl.LazyFrame, SerializedSchema | None]:
         source = kwargs.pop("source")
         lf = pl.scan_parquet(source, **kwargs)
-        metadata = _read_serialized_schema(source)
+        metadata = _read_serialized_schema(source, **_metadata_read_options(kwargs))
         return lf, metadata
 
     def read_frame(self, **kwargs: Any) -> tuple[pl.DataFrame, SerializedSchema | None]:
         source = kwargs.pop("source")
         df = pl.read_parquet(source, **kwargs)
-        metadata = _read_serialized_schema(source)
+        metadata = _read_serialized_schema(source, **_metadata_read_options(kwargs))
         return df, metadata
 
     # ------------------------------ Collections ---------------------------------------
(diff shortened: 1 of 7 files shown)


tests/failure_info/test_storage.py
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) QuantCo 2025-2026
 # SPDX-License-Identifier: BSD-3-Clause
 
+import uuid
+
 import polars as pl
 import pytest
 from fsspec import AbstractFileSystem, url_to_fs
```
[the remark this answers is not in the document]
mattijsdp: It's imported below

dataframely/_storage/parquet.py
```diff
@@ -245,11 +250,24 @@ def scan_failure_info(
...
-    meta = pl.read_parquet_metadata(path)
+def _metadata_read_options(kwargs: dict[str, Any]) -> dict[str, Any]:
+    # Forward only the options that `read_parquet_metadata` accepts (it has a narrower
+    # signature than `read_parquet`/`scan_parquet`) so the metadata read reaches the
+    # same store as the data read.
+    return {
+        key: kwargs[key]
+        for key in ("storage_options", "credential_provider", "retries")
+        if key in kwargs
+    }
```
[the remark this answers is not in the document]
AndreasAlbertQC: This seems sensible to me. WDYT @mattijsdp ?

AndreasAlbertQC (commented):
Thanks @mattijsdp, looks great! What do you think about the `inspect`-based suggestion by copilot? It seems like a good idea to me.

dataframely/_storage/parquet.py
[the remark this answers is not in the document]
mattijsdp: Sure, have added!

AndreasAlbertQC (approved):
Thanks! Looks good to me assuming CI goes through :) @borchero did you also want to review this?

dataframely/_storage/parquet.py
```diff
@@ -245,11 +251,23 @@ def scan_failure_info(
         return lf, serialized_rules, serialized_schema
 
 
-def _read_serialized_collection(path: str) -> SerializedCollection | None:
-    meta = pl.read_parquet_metadata(path)
+# `read_parquet_metadata` has no `**kwargs`, so unrecognized keys raise `TypeError`.
```
borchero: I don't really get this comment, let's just remove it


```suggestion
```

tests/conftest.py
```diff
@@ -44,6 +44,34 @@ def s3_tmp_path(s3_server: str, s3_bucket: str, monkeypatch: pytest.MonkeyPatch)
...
+    for var in (
+        "AWS_ENDPOINT_URL",
+        "AWS_ACCESS_KEY_ID",
+        "AWS_SECRET_ACCESS_KEY",
+        "AWS_ALLOW_HTTP",
+        "AWS_S3_ALLOW_UNSAFE_RENAME",
+        "AWS_DEFAULT_REGION",
+        "AWS_REGION",
+    ):
+        monkeypatch.delenv(var, raising=False)
```
borchero, line 69: I'm not sure this has the desired effect. Polars likely only reads these variables at startup and does not re-read them for every read. Could we check that?

I don't have a good alternative idea but I think all of those tests are moot if I'm right.

dataframely/_storage/parquet.py
mattijsdp: Sure, removed it. It's meant to explain why we don't just pass the kwargs as we do with `pl.scan_parquet/read_parquet`.

tests/conftest.py
mattijsdp: Good point! So yes, the tests were indeed moot as Polars caches an object store per bucket which isn't cleared by this monkeypatch. Other tests use the same bucket so it stayed reachable even without `storage_options`. This is per bucket though so I fixed this by creating a fresh bucket that is never env configured hence no cache and so requires `storage_options` forwarding. Ran the tests without the fix which then failed as expected. Let me know what you think.

tests/conftest.py
```diff
@@ -63,7 +67,11 @@ def s3_storage_options(
         "AWS_REGION",
     ):
         monkeypatch.delenv(var, raising=False)
-    return {
+    bucket = f"isolated-{uuid.uuid4()}"
+    boto3.client(
+        "s3", endpoint_url=s3_server, aws_access_key_id="", aws_secret_access_key=""
+    ).create_bucket(Bucket=bucket)
```
borchero: This will keep the bucket around indefinitely which is not ideal for local development & debugging. Could you `yield` from the fixture and delete the bucket after the `yield` statement as "teardown" operation?

borchero (commented):
Thanks for the update! 🚀 just one more comment :)

tests/conftest.py
mattijsdp: Sure, done.

Note: I didn't use a paginator and technically `list_objects_v2` caps at 1000 objects but I think this is fine for this use case?

mattijsdp:
Hey, is there anything outstanding for this PR. Would it be possible to merge?

borchero:
Hey @mattijsdp, we have decided that we will get rid of the IO metadata (see #367). Hence, I think this PR will not be necessary.

PR #356, opened 2026-06-26
fix: Explicitly specify `empty_as_null` for `explode`

borchero:
# Motivation

This fixes nightly CI: https://github.com/Quantco/dataframely/actions/runs/28209348571/job/83566986568.


AndreasAlbertQC (approved):

borchero:
Ah `polars-minimal` fails. @AndreasAlbertQC should we bump or how should we handle this?

AndreasAlbertQC:
Would be a shame to give up compat with older versions over this. I feel like we just have to bite the bullet and pass / not pass the argument based on `polars.__version__`

PR #359, opened 2026-07-05
feat: Add `Collection.pipe` method analogous to polars Lazy-/DataFrames

MoritzPotthoffQC:
# Motivation

I recently wanted to use this and realized I can't. I think it would be nice to support this so that we can use the same operation chaining pattern we use for data frames also for collections.

# Changes

- Add `Collection.pipe` method with the same interface as `pl.LazyFrame.pipe` (see [here](https://github.com/pola-rs/polars/blob/main/py-polars/src/polars/lazyframe/frame.py#L919))
- Add simple tests

### DIFF
--- a/dataframely/collection/collection.py
+++ b/dataframely/collection/collection.py
@@ -8,11 +8,21 @@ import sys
 import textwrap
 import warnings
 from abc import ABC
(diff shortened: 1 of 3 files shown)


borchero (approved):
Nice :)

PR #363, opened 2026-07-06
perf: Leverage `collect_all` for eager collection validation & filtering

borchero:
# Motivation

Using `collect_all` allows for considerable performance gains, esp. if the number of filters and/or members is large (and a sufficient number of CPU cores is available).

On my machine, this yields speedups of ~25-30% on our benchmark data.


AndreasAlbertQC (approved):
very nice!

PR #364, opened 2026-07-06
feat: Allow specifying the engine for `validate` and `filter`

borchero:
# Motivation

The streaming engine is sometimes faster; this speedup should accessible to the user.

In our benchmark, specifying `engine="streaming"` results in slowdowns of ~30-40% because of `maintain_order="left"` in our joins (which is ~2.5x faster on the in-memory engine compared to the streaming engine).

### DIFF
--- a/dataframely/collection/collection.py
+++ b/dataframely/collection/collection.py
@@ -406,8 +406,8 @@ class Collection(BaseCollection, ABC):
                 :meth:`~polars.LazyFrame.collect` on the individual member or
                 :meth:`collect_all` on the collection. Note that, in the latter case,
                 information from error messages is limited.
-            kwargs: Keyword arguments passed directly to :meth:`polars.collect_all` when
-                `eager=True`.
+            kwargs: Keyword arguments passed directly to :meth:`polars.collect_all` and
+                :meth:`polars.LazyFrame.collect` when `eager=True`.
 
         Raises:
             ValueError: If an insufficient set of input data frames is provided, i.e. if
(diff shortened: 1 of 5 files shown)


AndreasAlbertQC (approved):

delsner (approved):

dataframely/collection/collection.py
```diff
@@ -489,15 +497,19 @@ def validate(
...
+    ) -> bool:
         """Utility method to check whether :meth:`validate` raises an exception.
 
         Args:
             data: The members of the collection which ought to be validated. The
                 dictionary must contain exactly one entry per member with the name of
                 the member as key.
             cast: Whether columns with a wrong data type in the member data frame are
-                cast to their schemas' defined data types if possible.
+                cast to their schemas' defined data types if possible.]
```
ElisabethBrockhausQC:
```suggestion
                cast to their schemas' defined data types if possible.
```

dataframely/collection/collection.py
```diff
@@ -635,19 +658,19 @@ class HospitalInvoiceData(dy.Collection):
                         filter_keep.lazy().with_columns(pl.lit(True).alias(name)),
                         on=primary_key,
                         how="left",
-                        maintain_order="left",
+                        # maintain_order="left",
```
ElisabethBrockhausQC: If this has such an impact on performance, it would be nice if the user could disable it somehow if not needed. But I guess you don't want to extend the interface?

borchero: Yeah, fair point. I've been contemplating adding a `maintain_order: bool = True` parameter but didn't see any performance gain on real data (yet), so, I decided against it for now 😅

PR #370, opened 2026-07-21
fix: Ignore failure propagation in collection member that is ignored in filters and reject failure propagation without common primary key

MoritzPotthoffQC:
# Motivation

Opposed to the docs, for a member that is ignored in filters, the failure propagation property is currently not ignored.  

Additionally, we should reject collections if they have a failure-propagating member (not ignored) but no common primary key. Like for filters, entries that are filtered out of a failure-propagating member would be propagated across members, which does not work without common primary key. 

# Changes

- Ignore failure propagation for members that are ignored in filters
- Extend the criteria for rejection
- Add tests


borchero (approved):
Thank you!

PR #372, opened 2026-07-23
refactor!: Replace `eager: bool` with "lazy-in-lazy-out"

borchero:
# Motivation

Currently, we have a parameter `eager: bool` to govern whether validation and filtering is performed eagerly or lazily. This is a little inconsistent with what one would expect when familiar with the polars API:

- For schema validation/filtering, we can easily replace the parameter entirely with "lazy-in-lazy-out"/"eager-in-eager-out" behavior. This also makes user code more consistent as collection is not implicit in the case of a lazy input, but requires an explicit `collect`.
- For collection validation/filtering, we take the same approach as for `sink_parquet`: by default, we `collect`, but this can be "disabled" by setting `lazy=True` (unless the collection defines at least one eager member).


delsner (approved):
Nice!

PR #374, opened 2026-07-25
fix: Evaluate simple rules before joining group rule columns

### LINKED ISSUE #332: Rules using dynamic column selection can yield unexpected results
Inspired by https://github.com/Quantco/dataframely/discussions/331#discussioncomment-16654745

Here's an example with a rule that relies on dynamically selecting all boolean columns.
If there are no group rules, this works as expected. 
Adding a group rule can make validation fail unexpectedly. I have not debugged this in detail, but I thikn the reason is that we chain `with_columns` [here](https://github.com/Quantco/dataframely/blob/main/dataframely/_rule.py#L224C17-L226) in a way that makes the rule expression "see" the result of the group rule evaluation.

MWE:

```python
import polars as pl

import dataframely as dy


class MySchema(dy.Schema):
    x = dy.Bool()

    @dy.rule()
    def myrule(cls):
        return ~pl.any_horizontal(pl.col(pl.Boolean))


class MySchemaWithGroupRule(MySchema):
    @dy.rule(group_by="x")
    def my_group_rule(cls):
        return pl.lit(True)


df = pl.DataFrame({"x": [False]})

# Succeeds
MySchema.validate(df)

# Fails
MySchemaWithGroupRule.validate(df)
```

Result:

```bash
  File "/Users/x/repos/dataframely/test.py", line 26, in <module>
    MySchemaWithGroupRule.validate(df)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^
  File "/Users/x/repos/dataframely/dataframely/schema.py", line 573, in validate
    raise ValidationError(
        format_rule_failures(list(failure.counts().items()))
    )
dataframely.exc.ValidationError: 1 rules failed validation:
 - 'myrule' failed for 1 rows
```

Expectation: validation should succeed in both cases.

Sanjays2402:
Closes #332

# Motivation

A rule using dynamic column selection can fail unexpectedly as soon as the schema also has a group rule. In the reported MWE, `~pl.any_horizontal(pl.col(pl.Boolean))` validates fine on its own but fails once a `@dy.rule(group_by="x")` is added.

`with_evaluation_rules` joined the group rule evaluation columns onto the frame *first* and only then evaluated the simple rule expressions. Group rule columns are boolean, so the dynamic selector also picked up those internal columns.

# Changes

Simple expressions are now added before the group rules are joined, so they only ever see the user's own columns. Group rule evaluation is unaffected because those rules are computed from the original frame either way.

Adds `test_dynamic_column_selection_with_group_rule`, which fails on `main` and passes with this change.

### DIFF
--- a/dataframely/_rule.py
+++ b/dataframely/_rule.py
@@ -218,12 +218,15 @@ def with_evaluation_rules(lf: pl.LazyFrame, rules: dict[str, Rule]) -> pl.LazyFr
 
     # Before we can select all of the simple expressions, we need to turn the
     # group rules into something to use in a `select` statement as well.
+    # NOTE: The simple expressions are evaluated BEFORE the group rules are joined onto
+    #  the frame. Otherwise, rules using dynamic column selection (e.g.
+    #  `pl.col(pl.Boolean)`) would also select the boolean group rule columns.
     result = (
         # NOTE: A value of `null` always validates successfully as nullability should
         #  already be checked via dedicated rules.
-        lf.pipe(_with_group_rules, group_rules).with_columns(
+        lf.with_columns(
             **{name: expr.fill_null(True) for name, expr in simple_exprs.items()},
-        )
+        ).pipe(_with_group_rules, group_rules)
     )
 
     # If there is at least one rule that checks for successful dtype casting, we need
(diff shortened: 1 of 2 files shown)


borchero (commented):
Thank you for tackling this @Sanjays2402! I'm wondering whether this is the most ergonomic way to solve it: an alternative option would be to evaluate group rules via `over()` instead of `group_by+join` which would also potentially result in speedups. Could you try this out and benchmark before and after using the existing benchmark tests (by running `pixi run test-bench`)?

borchero:
This is superseded by #389.

PR #400, opened 2026-09-06
feat: Allow FailureInfo to persist only failing rule columns

### LINKED ISSUE #296: Feature Request: Add an argument to `FailureInfo.write_parquet(only_invalid_rules: bool = False)`
When dealing with very large schemas (e.g., 50 string columns, all with min and max lengths), the failure info parquet file ends up very large/wide. It's challenging to open, and challenging to sort horizontally through.

It'd be helpful if there was a way to only write the debugging rule output columns that were responsible for at least one failing row (i.e., only the columns with min length violations).

All the core columns from the dataframe being validated would still be included - this feature request is only scoped to the additional columns like `id|min_length`, `id|max_length`, etc.

Danila-Pechenev:
# Motivation

Fixes #296.

Persisted `FailureInfo` objects can become very wide because they include an output column for every validation rule, even when most rules never fail.

# Changes

- Add the keyword-only `only_invalid_rules` option to `FailureInfo.write_parquet()`.
- When enabled, persist all data columns but only rule columns containing at least one `False`.
- Keep the parquet metadata consistent with the retained rule columns.
- Preserve existing lossless behavior by default.
- Support round-tripping reduced empty failure information.
- Document that the reduced representation intentionally omits successful and unknown-only rule outcomes.

# Testing

- Added coverage for failing, successful, unknown, and empty rule columns.
- Added eager-read and lazy-scan round-trip coverage.
- Added repeated reduced-serialization coverage.
- Verified counts and co-occurrence counts remain unchanged.


dataframely/filter_result.py
```diff
@@ -128,6 +128,9 @@ def details(self) -> pl.DataFrame:
         filters in addition to member-level rules, or when calling :meth:`Schema.filter`
         with `cast=True` and dtype-casting fails for a value.
         """
+        if len(self._rule_columns) == 0:
+            return self.invalid()
+
```
borchero, line 133: Why is this necessary?

dataframely/filter_result.py
```diff
@@ -166,24 +169,45 @@ def __len__(self) -> int:
...
         """Write the failure info to a single parquet file.
 
         Writes the invalid rows along with additional boolean rule columns indicating
         which validation rules failed. Unlike :meth:`invalid`, this includes columns
-        for each rule, where ``False`` indicates the rule failed for that row.
+        for each rule by default, where ``False`` indicates the rule failed for that row.
+
+        Setting ``only_invalid_rules`` produces a reduced, intentionally lossy
+        representation that omits rule columns containing only successful or unknown
+        outcomes.
```
borchero: The argument is already documented below

```suggestion
```

dataframely/filter_result.py
```diff
@@ -166,24 +169,45 @@ def __len__(self) -> int:
...
 
         Args:
             file: The file path or writable file-like object to which to write the
                 parquet file.
+            only_invalid_rules: Whether to write only rule columns containing at least
```
borchero: What about

```suggestion
            only_failing_rules: Whether to write only rule columns containing at least
```

borchero (commented):
Thank you for the contribution @Danila-Pechenev! Just some small suggestions

dataframely/filter_result.py
Danila-Pechenev: This is needed for an empty `FailureInfo` persisted with `only_failing_rules=True`. Since no rule contains a failure, the reduced representation has `rule_columns=[]`. After reading/scanning it back, `details()` would otherwise call `pl.col()` and raise a `TypeError`. Returning `invalid()` makes `details()` work for that reduced state as well.

borchero (approved):
Thank you! :)

